PR #18, opened 2024-01-31
Set up basic framework for network request management with offline support

pkalita-lbl:
Fixes https://github.com/microbiomedata/issues/issues/563

This is a pretty big PR, but I think this is the minimum set of coherent changes. I'm happy to walk through this in a meeting if that would be helpful, but here's a rough outline of what's going on here.

### New dependencies
* Everything in the `@tanstack` namespace is related to adding [TanStack Query](https://tanstack.com/query/latest/docs/framework/react/overview) and some associated add-ons. These are the main packages that will underpin our server data management layer. This package _does not_ make server requests for you. Rather, it manages the state of those requests and a cache of the results. **A quick skim through their documentation is highly recommended before going much further!**
* [`@ionic/storage`](https://github.com/ionic-team/ionic-storage) is added to facilitate persistent storage of app data. Currently it is connected to TanStack Query's persistence provider. In the future we can also use it for saving any other data like user preferences.
* [`msw`](https://mswjs.io/docs/) is a API request mocking library. It is used in tests so that we don't need to rely on a live server.
* [`immer`](https://immerjs.github.io/immer/) is just a great utility library that simplifies working with immutable data.

### The submission hooks
The core of what's being added here is in `src/queries.ts`. This module exports two React hooks for working with `nmdc-server` data: `useSubmissionList` and `useSubmission`. 

The `useSubmissionList` hook is just a simple wrapper around TanStack Query's `useQuery` hook. Since the underlying API request is paginated, this hook accepts an object representing pagination options. This hook would be used by a component that needs to display a list of submissions to a user.

The `useSubmission` hook is a bit more involved. The idea is that you provide this hook the ID of a submission and it returns the tools you need to both get (the `query`) and update (the `mutation`) that submission. To accomplish that, It wraps _two_ hooks provided by TanStack Query: `useQuery` and `useMutation`. 

When we fetch a list of submissions, the response contains a list of _full_ submission objects. Therefore when we request a single submission, we want to check to see if the cached data from the list request has the submission we're interested in. If it's there we can present that to the user first before fetching (in the background) the single submission. This is the purpose of the `initialData` option provided to `useQuery`. 

When we update (a.k.a mutate) a submission we want a few things to happen. Before the request goes out (`onMutate`) we see if any mutations with the same `mutationKey` are paused and drop all but the latest one. This would happen if multiple mutation requests were issued while offline. Second, we grab the data before the mutation which will be used in case the mutation fails, and finally we do an optimistic update of the data in the cache. If the request goes out and fails (`onError`) we rollback the optimistic update using the data from before the mutation. It might look a bit odd at first that the `mutationFn` is not defined in this use of `useMutation`. We instead use the `queryClient.setMutationDefaults` function. This is required in order to resume persisted mutations, see: https://tanstack.com/query/latest/docs/framework/react/guides/mutations#persisting-offline-mutations. 

The last thing I'd point out in `src/queries.ts` is that everything TanStack Query does is based around keys (`queryKey` or `mutationKey`). The `submissionKeys` object holds functions to produce those keys in a consistent way. See: https://tkdodo.eu/blog/effective-react-query-keys.

### The API request functions
As mentioned before, TanStack Query does not make API requests. You provide it with functions that do those requests. The `src/api.ts` module defines those functions for us. Currently it sends the cookie-based credentials (`credentials: 'include',`) that I've been using as a stop-gap measure to talk to a local `nmdc-server` API. Eventually we can replace that with the real authentication mechanism. This module also defines a number of TypeScript interfaces to describe the data exchanged with the `nmdc-server` API. These types are _fully_ fleshed out, but it's enough to get started. 

### Testing
The infrastructure to support testing these new features is in `src/mocks`. The `src/mocks/fixtures.ts` module defines a list of mock submissions. The `src/mocks/server.ts` uses the `msw` library to define a few [mock request handlers](https://mswjs.io/docs/basics/intercepting-requests#http-requests) and a [server](https://mswjs.io/docs/api/setup-server/) to do the actual request interception. The request handlers return data defined in `src/mocks/fixtures.ts`. The server is enabled, reset, and disabled in `src/setupTests.ts`.

The tests in `src/api.test.ts` are relatively straightforward. They call the network request functions and verify that the data returned is what is expected based on the fixture data. The tests in `src/queries.test.tsx` are a bit more involved, but they're mainly built around the ideas in https://tanstack.com/query/latest/docs/framework/react/guides/testing. 

I updated the `test.yml` workflow file to replace the `$default-branch` variable (which is only available in [starter workflows](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization)) with our actual default branch name in order to get the tests running on PRs.

### Caveats
You'll see a lot of `TODO` comments referring to not hardcoding pagination options in various places. The gist here is that the pagination options are part of the submission list query key. So when you want to interact with that data you can't just say "give me the submission list data" you have to say what _page_ of data you're talking about. In the future we will have to keep track of what was the last page of results the user looked at, or we could also look into using [infinite queries](https://tanstack.com/query/latest/docs/framework/react/guides/infinite-queries). Either way I think that will come about when more of the actual UI is built out. And at that point we can decide what the best way to remove those hardcoded pagination options would be.


src/App.tsx
```diff
@@ -33,25 +40,62 @@ export const PATHS = {
   WELCOME_PAGE: "/welcome",
 };
 
+const queryClient = new QueryClient({
```
eecavanna, line 43: I see the documentation for this constructor at: https://tanstack.com/query/latest/docs/reference/QueryClient

The [network mode docs](https://tanstack.com/query/latest/docs/framework/react/guides/network-mode#network-mode-online) say the default network mode is `online`.

> Since TanStack Query is most often used for data fetching in combination with data fetching libraries, the default network mode is [online](https://tanstack.com/query/latest/docs/framework/react/guides/network-mode#network-mode-online).

src/App.tsx
```diff
@@ -33,25 +40,62 @@ export const PATHS = {
...
+      networkMode: 'online',
+      staleTime: 1000 * 20, // 20 seconds
+      gcTime: 1000 * 60 * 60 * 24 * 7, // 1 week
+      retry: 0,
+    }
+  }
+});
+addDefaultMutationFns(queryClient);
+
+const store = new Storage();
```
eecavanna, line 58: The [README.md file](https://github.com/ionic-team/ionic-storage) of `@ionic/storage` says:

> A simple key-value Storage module for Ionic apps. This utility uses the best storage engine available on the platform without having to interact with it directly (some configuration required, see docs below).

src/App.tsx
```diff
@@ -33,25 +40,62 @@ export const PATHS = {
...
-        <Route exact path={PATHS.HOME_PAGE}>
-          <Home />
-        </Route>
-        <Route exact path="/">
-          <Redirect to={PATHS.WELCOME_PAGE} />
-        </Route>
-      </IonRouterOutlet>
-    </IonReactRouter>
-  </IonApp>
+  <PersistQueryClientProvider
```
eecavanna, line 70: The documentation for the `PersistQueryClientProvider` component is at: https://tanstack.com/query/latest/docs/framework/react/plugins/persistQueryClient#persistqueryclientprovider

src/App.tsx
```diff
@@ -33,25 +40,62 @@ export const PATHS = {
...
+          <Route exact path={PATHS.HOME_PAGE}>
+            <Home />
+          </Route>
+          <Route exact path="/">
+            <Redirect to={PATHS.WELCOME_PAGE} />
+          </Route>
+        </IonRouterOutlet>
+      </IonReactRouter>
+    </IonApp>
+    <ReactQueryDevtools initialIsOpen={false} />
```
eecavanna, line 97: The docs for the dev tools are here: https://tanstack.com/query/latest/docs/framework/react/devtools

Specifically, the docs say:

> Devtools are excluded in production builds.

src/api.ts
```diff
@@ -0,0 +1,105 @@
...
+  piEmail: string;
+  piOrcid: string;
+  linkOutWebpage: string[];
+  studyDate: Nullable<string>;
+  description: string;
+  notes?: string;
+  contributors: Contributor[];
+}
+
+// TODO: replace the `object` types with the actual types
```
eecavanna, line 24: Thanks for including this `TODO` comment. Do you know where the actual type information would come from, if someone were to want to do this replacement? For example, is there a specific file (or directory) in another repository, that the authoritative type information is currently in?

src/global.d.ts
```diff
@@ -0,0 +1 @@
+type Nullable<T> = T | null | undefined;
```
eecavanna, line 1: Looks to me like this is a custom type that can be used anywhere in the code base (for example, [here](https://github.com/microbiomedata/nmdc-field-notes/pull/18/files#diff-769911c416ccf8514d8fd941ae0abe8fb5c606ade0c218e22151a5f5f9f3d700R18)) without having to `import` it there.

src/mocks/fixtures.ts
```diff
@@ -0,0 +1,152 @@
+import {SubmissionMetadata} from "../api";
+
+export const submissions: SubmissionMetadata[] = [
+  {
```
eecavanna, line 4: @pkalita-lbl, will you add a comment about how you generated this example object / how someone might be able to generate an example object in the future?

src/mocks/server.ts
```diff
@@ -0,0 +1,63 @@
...
+  async () => {
+      // This delay helps tests validate optimistic updates before the request fails
+      await delay(300)
+      return new HttpResponse(null, { status: 500 })
+  }
+)
+
+export const server = setupServer(...handlers);
+
+export { delay }
```
eecavanna, line 63: @pkalita-lbl, will you format these new files with Prettier before merging them into `main`?

src/queries.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+import {act, renderHook, waitFor} from "@testing-library/react";
+import {addDefaultMutationFns, useSubmission, useSubmissionList} from "./queries";
+import {produce} from "immer";
+import {patchMetadataSubmissionError, server, delay} from "./mocks/server";
+
+interface TestWrapperProps {
+  children: ReactNode
+}
+const createWrapper = () => {
+  const queryClient = new QueryClient({
```
eecavanna, line 16: The docs for `QueryClient` are here: https://tanstack.com/query/latest/docs/reference/QueryClient

src/queries.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+  await waitFor(() => expect(result.current.query.isSuccess).toBe(true));
+  expect(result.current.query.data?.id).toBe(TEST_ID_1)
+});
+
+
+test('useSubmission should return initial data from the list query', async () => {
+  const wrapper = createWrapper();
+
+  // First fetch data from the submission list query
+  // TODO: fix hardcoded pagination options
```
eecavanna, line 58: Leaving this as a `TODO` works for me. You explained this yesterday.

src/queries.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+  await waitFor(() => expect(listResult.current.isSuccess).toBe(true));
+
+  // Second fetch data from the individual submission query. Wait for the background fetch to
+  // complete before proceeding.
+  const { result } = renderHook(() => useSubmission(TEST_ID_1), { wrapper })
+  await waitFor(() => expect(result.current.query.isFetching).toBe(false));
+
+  // Update the submission data. There doesn't seem to be a good way to ensure that the onMutate()
+  // function is called without waiting for the entire mutation to complete. To verify that the
+  // data was updated optimistically, we initiate the mutation and then introduce a very short delay
```
eecavanna, line 89: Note to myself (@eecavanna): Look into this after I become more familiar with React Query.

src/queries.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+  // Once the mutation completes (as failed), ensure that both the mutation, query, and list query
+  // now return the rolled back data
+  await waitFor(() => expect(result.current.mutation.isError).toBe(true));
+  expect(result.current.query.data?.metadata_submission.studyForm.studyName).toBe('TEST 1');
+
+  // The list should also have the rolled-back data
+  const submissionFromList = listResult.current.data?.results
+    .find((submission) => submission.id === TEST_ID_1);
+  expect(submissionFromList?.metadata_submission.studyForm.studyName).toBe('TEST 1');
+})
```
eecavanna, line 203: Thank you for adding tests!

src/App.tsx
yxu-lanl, line 1: Which storage do we want to use? (LocalStorage, IndexedDB, or SQLite)

src/queries.ts
yxu-lanl, line 1: const options: PaginationOptions = { limit: 10, offset: 0 } // TODO: can't hardcode this, should come from some state

Do you mean we need a state management? e.g., Redux, MobX...

yxu-lanl (approved):

src/App.tsx
pkalita-lbl: I added https://github.com/ionic-team/ionic-storage which acts as an interface to both `IndexedDB` and `localstorage`, choosing between them based on what's available. I believe that by default it prefers `IndexedDB` but that can be [configured](https://github.com/ionic-team/ionic-storage?tab=readme-ov-file#configuration). It can also be [connected to a SQLite store](https://github.com/ionic-team/ionic-storage?tab=readme-ov-file#sqlite-installation). I didn't see the need for that at the moment, but if we do find we need it in the future it seems like it would be easy to integrate.

src/queries.ts
pkalita-lbl: Possibly. Although having used Redux in the past I might shy away from it unless there was a _really_ compelling reason for it -- it's pretty complex and heavyweight. I don't have direct experience with MobX so I should probably learn more about it. What I actually was thinking was just directly using React [context](https://react.dev/reference/react/createContext). But if that's _too_ lightweight we can look at other options.

PR #19, opened 2024-02-02
Add workflow for deploying to GitHub Pages

pkalita-lbl:
Adding a simple workflow for building a deploying the website version of the app to GitHub Pages. I tend to like the [actions-based approach](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow) (even though it's technically still in beta). I'm guessing this repo's Settings are set to the branch-based deployment, so we should change that just before merging this.

Later we should think about releases and building from tags, but while we're still in early days I think building from `main` is okay.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,47 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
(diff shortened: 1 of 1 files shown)


eecavanna:
Looks good to me. One thing I want to keep in mind is that this would make it so the "deployed" web app is in sync with the tip of the `main` branch, which may not exactly align with what has been "Released" to the App Store or Play Store.

pkalita-lbl:
> Looks good to me. One thing I want to keep in mind is that this would make it so the "deployed" web app is in sync with the tip of the `main` branch, which may not exactly align with what has been "Released" to the App Store or Play Store.

Yes absolutely, we'll revisit this once we have a clearer picture of the release process.

Issue #20, opened 2024-02-03
Web: Server-side routing to non-root paths is broken


eecavanna:
Example:
- Click https://fieldnotes.microbiomedata.org/ -> page loads OK 🟢 
- Click https://fieldnotes.microbiomedata.org/welcome -> error 404 page 🔴 
- Click https://fieldnotes.microbiomedata.org/ and get redirected (via **client-side routing**) to https://fieldnotes.microbiomedata.org/welcome -> page loads OK 🟢 
    - Reload the page -> error 404 page 🔴 

I have experience with this issue from other projects. Here's a workaround I have used in the past: https://github.com/rafgraph/spa-github-pages

I also like the way the situation is summarized here: https://github.com/orgs/community/discussions/64096

PR #21, opened 2024-02-05
Implement `404.html` page that redirects to domain root

eecavanna:
Currently, when a client navigates to a non-root URL of the web app via a means other than client-side routing (e.g. via reloading the web page or via typing the URL into the address bar); GitHub Pages displays an HTTP 404 Error Page. This phenomenon is described in #20.

GitHub Pages supports the use of custom `404.html` pages, which is what I am introducing with this branch. This custom `404.html` page does one thing: it redirects the client to the root URL of the site, discarding any path or query string they might have originally requested. I consider this to be an improvement over the current behavior.

> Note: The reason I didn't use the hard-coded relative URL of `/` to refer to the root URL of the site; is that I wanted to lay a foundation for moving to a more sophisticated solution, such as the one referenced below.

Should we—in the future—decide to allow users to enter the app at non-root paths (i.e. to open the app directly to a "deep" screen), we can switch to a solution like the one described at https://github.com/rafgraph/spa-github-pages/.

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -1,0 +1,42 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>NMDC Field Notes</title>
+  </head>
+  <body>
+    <script>
+      /**
+       * Navigate the client to the domain root, discarding any path or query string
+       * present in the originally-requested URL.
+       *
+       * Note: This script was designed to work around the issue of GitHub Pages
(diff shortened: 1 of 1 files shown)


pkalita-lbl (approved):

yxu-lanl (approved):

eecavanna:
Thanks for reviewing this PR. I'll merge it into `main` now.

PR #23, opened 2024-02-05
Fix top level await error on build

pkalita-lbl:
This fixes the error that is happening when running `npm run build` (https://github.com/microbiomedata/nmdc-field-notes/actions/runs/7788383055/job/21237662321#step:5:32). I inadvertently introduced this error in #18. 

I also added a `npm run build` step to the main testing workflow so that we can catch this type of error in PRs instead of only when the deploy from `main` happens.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { Redirect, Route } from "react-router-dom";
 import { Storage } from "@ionic/storage";
 import { IonApp, IonRouterOutlet, setupIonicReact } from "@ionic/react";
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -55,47 +58,64 @@ const queryClient = new QueryClient({
...
-    getItem: async (key) => store.get(key),
-    setItem: async (key, value) => store.set(key, value),
-    removeItem: async (key) => store.remove(key),
-  },
-});
+  useEffect(() => {
+    async function initPersister() {
+      const store = new Storage();
+      await store.create();
+      setPersister(
```
eecavanna, line 68: Note to self: 

React components re-render when either (a) their props change or (b) their state changes. Since this `setPersister` call will change the component's state, it will cause a re-render; which will cause the...

```ts
  if (persister == null) {
    return null;
  }
```

...check to be skipped and for the "normal" element tree...

```ts
  return (
    <PersistQueryClientProvider
```

...to be rendered.

eecavanna (approved):

eecavanna:
The solution makes sense to me. Thanks for fixing this and updating the "test" workflow to ensure the app builds without errors.

yxu-lanl (approved):

PR #25, opened 2024-02-08
Replace boilerplate app name with real one

eecavanna:
### Summary of changes

- I updated the `index.html` and `public/manifest.json` files to say "NMDC Field Notes" (or "Field Notes") instead of "Ionic App"


yxu-lanl (approved):

eecavanna:
Thanks, @yxu-lanl !

Issue #27, opened 2024-02-09
Fix linking into token route on GitHub pages


pkalita-lbl:
After merging #26, I realized that the web version deployed on GitHub pages can't complete the login process successfully. The [GitHub Pages-deployed version](https://fieldnotes.microbiomedata.org/) is [configured](https://github.com/microbiomedata/nmdc-field-notes/blob/64e8b6b48b77d5a387a16772c76c9096a1f39521/.github/workflows/deploy.yml#L38) to use the [dev version](https://github.com/microbiomedata/nmdc-field-notes/blob/64e8b6b48b77d5a387a16772c76c9096a1f39521/.env.staging) of the portal. If you initiate the login flow, once you login with ORCID, the dev portal will successfully redirect you back to `https://fieldnotes.microbiomedata.org/token?...`. However since `/token` is a route defined within the React application itself, it triggers a 404 from GitHub Pages, and our [404.html page](https://github.com/microbiomedata/nmdc-field-notes/blob/main/public/404.html) just redirects you back to `/`.

@eecavanna already pointed out that this would be an issue in #21, but the connection to the authentication flow didn't occur to me at the time. So here are some things I think we should consider:

* As Eric pointed out: https://github.com/rafgraph/spa-github-pages 
* Another `404.html` based solution: https://www.smashingmagazine.com/2016/08/sghpa-single-page-app-hack-github-pages/
* We could also consider switching from `<IonReactRouter>` to `<IonReactHashRouter>` in `src/Router.tsx`. We would need to make an accompanying change to `nmdc-server` to redirect to `/#/token` instead of `/token`, but there would be no GitHub Pages trickery to do. I _think_ that mainly people shy away from hash routing because of aesthetic and SEO reasons, but those aren't really concerns for us.

eecavanna:
Thanks for summarizing the situation. The current behavior (problem) makes sense to me.

A fourth option I think we have is to create a `token.html` document alongside the existing `401.html` document. I think GitHub Pages would serve the former if someone requested `/token` (in contrast, I think `401.html` is a special "catch-all" document that GitHub will serve if no documents exist at the requested path). While I think this fourth option would work, I don't think it would scale well if we were to end up wanting to do it for other deep link URLs in the future.

pkalita-lbl:
[...]

PR #31, opened 2024-02-11
Support deep linking on GitHub Pages

eecavanna:
### Summary of changes

- Implemented https://www.smashingmagazine.com/2016/08/sghpa-single-page-app-hack-github-pages/ (and added a check that the originally-requested URL and the site URL have the same origin; which is something `history.replaceState()` requires, but is not addressed in the article)
- Added a fallback route so that, when someone visits a path that doesn't exist in the app (such as `/foo`), the app redirects them to the root path

### DIFF
--- a/index.html
+++ b/index.html
@@ -25,6 +25,41 @@
   </head>
   <body>
     <div id="root"></div>
+    <script>
+      /**
+       * Retrieves the originally-requested URL, if any, from `sessionStorage` and navigates to it on the client side.
+       *
+       * Details: If `sessionStorage` contains an originally-requested URL that both (a) has the same origin as the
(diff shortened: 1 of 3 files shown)


eecavanna:
I am ready for this to be reviewed/merged (and, as an alternative, ready to discuss pursuing a different one of the options listed in https://github.com/microbiomedata/nmdc-field-notes/issues/27).

I haven't tested this on GitHub Pages yet. It hasn't been merged into `main` yet and that is when the repo would automatically deploy it to GitHub Pages.

pkalita-lbl (approved):

yxu-lanl (approved):

PR #32, opened 2024-02-17
Convert inline styles into (Ionic or ad hoc) CSS classes

eecavanna:
### Summary of changes

- Removed all inline `style` attributes from the `WelcomePage` component
- Instead, used CSS classes—either Ionic ones (if available) or ad hoc ones (if the former were not available)


eecavanna:
I'm ready for this to be merged in. I won't tag anyone for review yet since it's the weekend.

pkalita-lbl (approved):

PR #36, opened 2024-02-20
Add a basic study list to home page

### LINKED ISSUE #29: Add list of studies to home screen
Once a user is logged in, the home screen should display:

* Studies listed by name
* For each study an expandable list of samples by name
* A button/link to navigate to a page for creating a new study (the destination page doesn't need to be implemented with this issue)

Relevant part of the [mockups](https://design.penpot.app/#/workspace/c29b5282-da70-8012-8003-6e07eabfa764/aba078d3-8498-80bd-8003-70a9899d77ec?page-id=aba078d3-8498-80bd-8003-70a9899d77ed):
Untitled

Some things that may need discussion or clarification:
* I don't know if the "Sync with NMDC Submission Portal" button is necessary. I think that can happen automatically in the background.
* My recollection was that unstructured notes ("Quick Note" in the mockup) would be attached to specific studies. I'm not sure what the advantage of them being detached from a study would be -- it also has significant backend data storage implications to have them not attached to a study. 

Potentially useful Ionic component:
* https://ionicframework.com/docs/api/accordion

pkalita-lbl:
Fixes #29 

### Query hooks

I had added a `useSubmissionList` hook in a previous PR. At the time that hook used React Query's `useQuery` hook under the hood, and you passed `useSubmissionList` an object representing pagination parameters. These changes update `useSubmissionList` to use React Query's [`useInfiniteQuery`](https://tanstack.com/query/latest/docs/framework/react/reference/useInfiniteQuery) instead. This hook is indented to back an "infinite scroll" UI, which feels like a more natural, mobile-ish way to show the list of studies (as opposed to a traditional Previous/Next pagination) when the user has more studies than we get in the initial request. 

The key to `useInfiniteQuery` is that is stores _all_ of the pages of data it fetches in the `data.pages` field. This necessitated a few minor changes to the `useSubmission` hook where it gets or modifies data from the submission list cache. This also took care of the places where pagination options were hardcoded because those options are no longer part of the submission list query key.

### New StudyList component

The new StudyList component uses the `useSubmissionList` hook to fetch submissions and display them as a list of Studies. If a user has more than 10 studies associated with their account they will see a "Load More" button. Each study is displayed as the study name, the environment type, and the number of associated samples. Tapping on a study navigates to a study view page (just implemented a stub for now). The header of the list also includes a "New" link which navigates to a study create page (also just a stub for now). This is a bit different than what the mockups show, but I'm attempting to follow "standard" mobile UI patterns. We can always go back and tweak that type of thing later.

### Miscellany 

* The home page component was moved from `src/pages/Home.tsx` to `src/pages/HomePage/HomePage.tsx` to be inline with the filename and directory structure of other page components. 
* I must have run some `ionic` command that synced something in `ios/App/Podfile`. I suppose that probably should have been part of #26. Better late than never!

## Comments

As discussed in #29, Quick Notes will be associated with studies -- not standalone entities. So they don't need to be part of this display.

I looked into using https://ionicframework.com/docs/api/accordion, as the mockups suggest, but the resulting interface was actually a bit clunky. See thread on Slack from Feb 15th for further discussion.

The mockups also talk about adding an indicator for whether a study is/has been synced with the backend. That's still a good idea, but I think it makes more sense to fold that in later as study editing functionality is developed.

I did look at https://ionicframework.com/docs/api/infinite-scroll instead of a "Load More" button, but I found it to be actually a bit disappointing. It doesn't give you control over which element is use for scroll even listening (i.e. you can't put your list of studies in a scrolling container within the page). If the elements representing the first page fetch don't completely fill the `<IonContent>` container it doesn't fire its `ionInfinite` event until _after_ the use attempts to scroll. So you end with what looks like a half-loaded page until you start scrolling. In practice I don't think that many users will have more than 10 studies (we could even bump that default up to 20 if we wanted), so the number of people who will even see the "Load More" button is limited.


src/api.ts
```diff
@@ -29,7 +29,7 @@ export interface MetadataSubmission {
   templates: string[];
   studyForm: StudyForm;
   multiOmicsForm: object;
-  sampleData: object;
+  sampleData: Record<string, Record<string, string>[]>;
```
eecavanna, line 32: I don't know whether you got this TypeScript type definition from `nmdc-server`. I recently came to the conclusion that at least one of the type definitions currently in use in `nmdc-server` for a JSON value is inadequate. I filed an issue about that [here](https://github.com/microbiomedata/nmdc-server/issues/1142). 

I proposed this generic JSON value type instead:

```ts
/**
 * Define a TypeScript type alias that accounts for all possible JSON values.
 *
 * References:
 * - https://www.json.org/json-en.html
 * - https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/#more-recursive-type-aliases
 */
export type JSONValue = string | number | boolean | null | { [k: string]: JSONValue } | Array<JSONValue>;
```

src/components/Pluralize/Pluralize.tsx
```diff
@@ -0,0 +1,24 @@
...
+const Pluralize: React.FC<PluralizeProps> = ({
+  count,
+  singular,
+  plural,
+  showCount,
+}) => {
+  return (
+    <>
+      {showCount && <>{count} </>}
+      {count === 1 ? singular : plural || `${singular}s`}
```
eecavanna, line 19: This expression took me a few seconds to parse. I'd prefer the "fallback" behavior be defined separately, so as to simplify the expression:

```diff
+ const pluralStr = plural || `${singular}s`;
  return (
    <>
      {showCount && <>{count} </>}
-     {count === 1 ? singular : plural || `${singular}s`}
+     {count === 1 ? singular : pluralStr}
``

Comment (eecavanna) [reply src/components/Pluralize/Pluralize.tsx:19]: Also, here's a link to an [NPM package](https://github.com/plurals/pluralize) (plus TS [types](https://www.npmjs.com/package/@types/pluralize) for it) called "pluralize" that might come in handy for us in the future. I've never used it before. I think it was designed to handle some of the pluralization special cases in English.

File: src/queries.ts [hunk 3]
```diff
@@ -44,10 +40,29 @@ export function useCurrentUser() {
...
+export function useSubmissionList() {
+  return useInfiniteQuery({
+    queryKey: submissionKeys.list(),
+    queryFn: ({ pageParam }) =>
+      nmdcServerClient.getSubmissionList({
+        limit: PAGE_SIZE,
+        offset: pageParam * PAGE_SIZE,
+      }),
+    initialPageParam: 0,
+    getNextPageParam: (lastPage, _, lastPageParam) => {
```
eecavanna, line 54: I find the prefix `last` to be ambiguous here (i.e. whether it refers to the "previous" page or the "final" page).

eecavanna (approved):
Thanks for including the comprehensive and clear PR description. This looks good to me. I left a few comments—none are deal-breakers from my perspective. I will play with the routing locally now and report back.

eecavanna:
I noticed that, when I am on the "/home" page (i.e. the list of studies) and I reload the web page, the web browser seems to fetching all studies; not just those that would fit within the viewport. I was surprised to see this.

https://github.com/microbiomedata/nmdc-field-notes/assets/134325062/4a3a44d0-a52a-4f74-ba78-78e04ec75b65

src/Router.tsx
```diff
@@ -41,8 +45,15 @@ const Router: React.FC = () => {
         <Route exact path={PATHS.TUTORIAL_PAGE}>
           <TutorialPage />
         </Route>
+        <AuthRoute path={PATHS.STUDY_VIEW_PAGE}>
+          <StudyViewPage />
+        </AuthRoute>
+        {/* Not sure why this needs to come after the view route, but it only works that way */}
```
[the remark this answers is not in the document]
eecavanna: I reproduced, locally, the behavior you alluded to in the code comment. In other words, when the "create" route is above the "view" route, the "view" route is the one that the app follows and the app shows the string `create` as though it were the ID of a study.



It doesn't make sense to me. Your code comment matches my observations. I also tried replacing the `AuthRoute`s with regular `Route`s and that didn't seem to me to make any difference.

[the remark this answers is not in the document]
eecavanna: While playing around with this phenomenon locally; in an attempt to eliminate some variables, I came up with this example that doesn't use any custom components:

image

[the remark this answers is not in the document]
eecavanna: I found that, if I wrap the contents of the `<IonRouterOutlet>` element in a `<Switch>` element (where `Switch` is imported from `react-router-dom`), the routing **does behave the way I expect**.

```diff
- import {Redirect, Route} from "react-router-dom";
+ import {Redirect, Route, Switch} from "react-router-dom";

// ...

    <IonReactRouter>
      <IonRouterOutlet>
+       <Switch>

// ...

+       </Switch>
      </IonRouterOutlet>
    </IonReactRouter>
```

The Ionic [documentation](https://github.com/ionic-team/ionic-docs/blob/4df75fbf279f1e4901bad988a865a39a9d738c1b/docs/react/navigation.md?plain=1#L583-L585) says:

> Since `IonRouterOutlet` takes over the job in determining which routes get rendered, using a `Switch` from React Router has no effect when used inside of an `IonRouterOutlet`.

That is not consistent with what I am seeing (assuming, where it says "from React Router," they are not making a distinction between `react-router` and `react-router-dom`).

src/api.ts
pkalita-lbl: Technically it should adhere to the `SampleData` class defined in the submission-schema: https://github.com/microbiomedata/submission-schema/blob/fff9173fc4aa42a609fcb7e86b3f71c0a10d5128/src/nmdc_submission_schema/schema/nmdc_submission_schema.yaml#L51005-L51024. So it's a little more restricted than just any JSON value. This is sort of a half-step towards that. Once we get into sample editing we'll need to integrate the submission schema. At that point we could also look into [generating TypeScript](https://linkml.io/linkml/generators/typescript.html) from the LinkML schema.

src/components/Pluralize/Pluralize.tsx
pkalita-lbl: I took your suggestion on making the logic more clear. The extra space in `<>{count} </>` is intentional. Otherwise Reach will collapse the whitespace between the two expressions and you'd end up with "2Samples". I added a comment to explain that.

pkalita-lbl:
I believe that's actually the expected behavior. Once you've fetched, say, up to page 5, those 5 pages are now in React Query's cache. When you reload the page it returns all 5 pages from the cache and starts a background refetch. The docs explain that it does refetch each page sequentially: https://tanstack.com/query/latest/docs/framework/react/guides/infinite-queries#what-happens-when-an-infinite-query-needs-to-be-refetched It suggests that there are ways to manually modify the query cache to only refetch the first page. A little further on it explains that you can also mitigate the multiple page refetch by limiting the number of pages it stores: https://tanstack.com/query/latest/docs/framework/react/guides/infinite-queries#what-if-i-want-to-limit-the-number-of-pages But you'd have to account for that in the UI by sometimes showing a "Load Previous" button. 

So yes, I saw the same thing you saw. I guess I convinced myself that (a) it's expected behavior and (b) not that many people will even hit this scenario (and if we bumped up the page size to 20, even fewer) and (c) `nmdc-server` can handle it 😁. But if we decide it's something we want to address I think there are avenues to explore.

src/Router.tsx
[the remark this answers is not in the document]
eecavanna: I'm willing to proceed with it for now, too. I like that you pointed it out in the code comment. 

I will create an issue about it (not assigned to any person or sprint) for future reference.

src/api.ts
eecavanna: OK, thanks for clarifying. I'm OK with this being merged in as is, given that we will be revisiting it at a later stage of development.

eecavanna:
Thanks for addressing my concerns. I'm OK with this branch being merged in as is. 

I'm still not clear on the "last" prefix (i.e. "last" as in "previous" versus "last" as in "final"). I'm OK with it being merged in and addressed later, though.

PR #37, opened 2024-02-23
Study view page

### LINKED ISSUE #30: Add route for viewing an individual study
When tapping on a study listed on the home screen, the user should go to a route for that individual study. If I'm interpreting the [mockups](https://design.penpot.app/#/workspace/c29b5282-da70-8012-8003-6e07eabfa764/aba078d3-8498-80bd-8003-70a9899d77ec?page-id=aba078d3-8498-80bd-8003-70a9899d77ed) correct I think the most important parts are:

* A condensed view of study-level information (name, description, perhaps environment type)
* A list of samples with an affordance to add a new sample
* (not represented in mockups but see discussion in #29) A list of quick notes and an affordance to add a new note
* A button/link to navigate to a route for editing the study-level information (a new route that doesn't need to be implemented with this issue)
* A button to delete the study

Relevant part of mocksup:
Image

Backend considerations:
* Study data is stored as a JSON blob in the portal Postgres database which means it should be easy enough to add new fields for storing things like notes without having to add database tables. But we should update any relevant [backend data models](https://github.com/microbiomedata/nmdc-server/blob/dda5699ca0585bc7e42a2c8b97991e9715f919cf/nmdc_server/schemas_submission.py) to make it clear where we're storing the data.
* I don't believe there is currently an endpoint for deleting a study, so we'll likely need to add that

Other thoughts:
* Similar to #29, the "Sync with NMDC Submission Portal" button shown in the mockups is probably unnecessary, but we might want to show some indication of whether the sync has happened in the background.
* The export feature was listed as a Nice to Have and not a Requirement for the MVP (https://docs.google.com/spreadsheets/d/17LnRDTAnX-IhDDTscGAt6yJASYAmX0pwbW3LkukbSW8/edit#gid=1463384098) so let's leave that off for now.

pkalita-lbl:
Fixes #30 

## Summary

These changes fill out the individual study view page, which was previously just a stub page.

### New dependencies

* [`minisearch`](https://github.com/lucaong/minisearch) and [`react-minisearch`](https://github.com/lucaong/react-minisearch) are added to support filtering the list of samples on the study view page. It may seem like overkill compared to a one-liner like `const searchResults = samples.filter(sample => sample.samp_name.includes(searchTerm)`, but (1) this library handles details like multi-term tokenization, token combination, case-sensitivity, etc better than we would from scratch; (2) I think we'll end up reusing these in other places (like the sample edit interface).

### Routing

* Previously `src/Router.tsx` exported an object of string constants representing the routes of various pages. I have updated this object to make some of the values functions instead. This is useful for paths that have a variable parameter in them. I also removed the "page" prefix from all the keys because it seemed a bit redundant. 
* _The order of routes continues to be a mystery_. After Eric seemed to get the "expected" definition order working in #38, when I merged those changes into this branch they weren't working 😧. I resorted to reverting to the "unexpected but works" order.
* I added new routes with stub pages for study editing, sample view/edit (I'm assuming that will be one route), and sample creation.

### New components

* `src/components/NoneOr/NoneOr.tsx` -- a simple utility component for displaying placeholder text if a given value doesn't exist. The intention is to reduce the boilerplate of doing things like `{object.field || <i>None</i>}` for a bunch of different fields. That's also where the admittedly weird name of the component came from. 
* `src/components/SectionHeader/SectionHeader.tsx` -- a presentational component for "minor" section headers.
* `src/components/SampleList/SampleList.tsx` -- a component for displaying the samples of a submission. This component does not fetch data; it expects the submission to be passed to it. By default it shows only the first 5 samples and, if there are more than 5 samples, shows a button to show the rest. It also uses an `<IonSearchbar>` component to filter the list of samples. Each individual sample links to a stub sample page. The list header links to a stub sample create page.
* `src/components/StudyView/StudyView.tsx` -- the main component displayed by the study view page. This component is responsible for fetching submission data (via the `useSubmission` hook). It displays study-level data as simple list entries and uses the `<SampleList>` component to display sample information.

### Note about sample IDs

**Samples don't have IDs**. They're essentially only identified by their index in the `sampleData.<template>_data` array. But because of the filtering functionality (and perhaps sorting one day) the _display_ index is not necessarily the same as the _storage_ index. To account for this I updated the data fetching functions in `src/api.ts` to inject a new field (`_index`) into the `sampleData.<template>_data` array elements that captures the original _storage_ index. This is later used in `src/components/SampleList/SampleList.tsx` for `key` props and generating routes.


src/components/SectionHeader/SectionHeader.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { HTMLAttributes, PropsWithChildren } from "react";
+import styles from "./SectionHeader.module.css";
+
+const SectionHeader: React.FC<HTMLAttributes<PropsWithChildren>> = ({
+  children,
+  className,
+}) => {
+  return (
+    <div className={`${styles.header} ion-margin ${className}`}>{children}</div>
```
eecavanna, line 9: Oh! Is this how scoping of styles works in this framework? I had been [prefixing the class names](https://github.com/microbiomedata/nmdc-field-notes/blob/main/src/pages/WelcomePage/WelcomePage.css) in an attempt to avoid styles designed for one component impacting other components. I didn't know the framework came with a ready-to-use scoping solution.

pkalita-lbl: It's not an Ionic-specific thing, so I'm not sure they even mention it in their docs. But Ionic's build tooling is based on around Vite, and [Vite supports CSS Modules](https://vitejs.dev/guide/features.html#css-modules) out of the box.

eecavanna: Thanks! I had both seen that syntax before (on another project) and come across that name ("CSS Modules") before, but didn't know they were the same thing, and also didn't know that was something we could use out of the box. I'll create an issue for myself to update the "legacy" CSS I wrote a few days ago, to use CSS Modules. Resolving this thread.

eecavanna:
I will start reviewing this PR this afternoon.

eecavanna:
I will start reviewing this now.

src/components/SampleList/SampleList.tsx
```diff
@@ -0,0 +1,159 @@
...
+            onClick={() => setCollapsed((previous) => !previous)}
+          >
+            {collapsed ? "Show All" : "Show Less"}
+          </IonButton>
+        )}
+
+        {displaySamples.length === 0 && (
+          <div className="ion-text-center">
+            <IonNote color="medium" className="ion-padding">
+              <i>{searchTerm ? "No samples match search" : "No samples yet"}</i>
```
eecavanna, line 150: This works for me. An alternative that came to my mind:
```diff
- "No samples match search" : "No samples yet"
+ "No matching samples" : "No samples yet"
```

src/components/StudyList/StudyList.tsx
```diff
@@ -37,7 +32,7 @@ const StudyList: React.FC = () => {
     <>
       <IonListHeader>
         <IonLabel>Studies</IonLabel>
-        <IonButton routerLink={PATHS.STUDY_CREATE_PAGE}>New</IonButton>
+        <IonButton routerLink={paths.studyCreate}>New</IonButton>
```
eecavanna, line 35: Since we also have "Edit" on the screen, I'm wondering whether using a verb here will result in people finding the UI more cohesive. The verb that first came to mind is "Create", but that seems long to me. The next one that came to mind is "Add". "New" definitely works for me for now—I don't want to hold up the merge for this.

src/pages/SamplePage/SamplePage.tsx
```diff
@@ -0,0 +1,40 @@
...
+} from "@ionic/react";
+import { useParams } from "react-router";
+import { paths } from "../../Router";
+
+interface SamplePageParams {
+  submissionId: string;
+  sampleIndex: string;
+}
+
+const SamplePage: React.FC = () => {
```
eecavanna, line 19: I was going to say I had already implemented a `PlaceholderPage`... then I realized this is referring to a biosample. I don't know that I would be in favor of renaming this to `BiosamplePage`, but I did want to record my initial confusion. I do want the user-facing UI to use the word "Sample" (I think it's a more user-friendly word and they'll know what it's referring to).

eecavanna (approved):
Looks great! I left some comments, most of which I will resolve right after leaving this review.

src/components/SampleList/SampleList.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <IonButton routerLink={paths.sampleCreate(submission.id)}>
+          New
+        </IonButton>
+      </IonListHeader>
+
+      <IonGrid className={showSearch ? "ion-hide" : ""}>
+        <IonRow class="ion-justify-content-between">
+          <IonCol size="auto">
+            <IonChip outline>
+              Template: {submission.metadata_submission.templates[0]}
```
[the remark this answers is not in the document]
eecavanna: According to the types, it can be an empty list.

https://github.com/microbiomedata/nmdc-field-notes/blob/c98d6ed89ce1c9b24983efbb4acfbd3fa8135593/src/api.ts#L29

Do you still want to show the chip in that situation?

src/components/SampleList/SampleList.tsx
```diff
@@ -0,0 +1,159 @@
...
+          </IonCol>
+          <IonCol size="auto">
+            <IonButton
+              title="show sample search"
+              size="small"
+              fill="clear"
+              onClick={() => {
+                setShowSearch(true);
+                // Wait for the search bar to be rendered before focusing
+                setTimeout(() => searchElement.current?.setFocus(), 10);
```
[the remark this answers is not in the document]
eecavanna: Here's a `useEffect` call that applies focus to the search field whenever the search field changes from hidden to visible. This could be used instead of the `setTimeout` call. This would go at the top level of the component (i.e. the same level as the `useState` calls).

```ts
  // Whenever the search field changes from hidden to visible, focus on it.
  useEffect(() => {
    if (showSearch) {
      searchElement.current?.setFocus();
    }
  }, [showSearch]);
```

eecavanna:
I'm done reviewing this and cleaning up my initial review comments.

src/components/SampleList/SampleList.tsx
eecavanna, line 85: According to the types alone, `templates` can be an empty list (for the record, I don't know of any way to constrain that via types alone).

https://github.com/microbiomedata/nmdc-field-notes/blob/c98d6ed89ce1c9b24983efbb4acfbd3fa8135593/src/api.ts#L29

I don't know whether you want the `<IonChip>` to be rendered that situation (i.e. when `templates` is an empty list).

src/api.ts
```diff
@@ -96,20 +101,45 @@ class NmdcServerClient extends FetchClient {
...
   }
 
+  /**
+   * Injects the "_index" property into each sample in the submission's sampleData.
+   *
+   * Because the submission portal backend needs to be tolerant of storing "invalid" data, some
+   * samples could essentially be empty. Therefore, there are no existing fields we can treat as a
+   * key (or persistent identifier). The samples are essentially identified by their position in the
+   * array. However, because the SampleList component allows the user to filter samples, we need to
+   * keep track of the original index of each sample.
```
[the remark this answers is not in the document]
pkalita-lbl: This gets to a bit of nomenclature that doesn't get discussed much. In user-facing language we talk about Studies and Samples. But as far as the backend is concerned a Study and its Samples are bundled together in one object called a `SubmissionMetadata` (don't ask me, I didn't name it). In order to get a list of Studies you need to use the API endpoint to get a list of `SubmissionMetadata` objects (which is paginated). But once you have an individual `SubmissionMetadata`, the Study information is there and _all_ of the Samples for that Study are there as well. So there's no pagination involved when dealing with the Samples of an individual Study.

src/components/SampleList/SampleList.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+const SampleList: React.FC<SampleListProps> = ({
+  submission,
+  collapsedSize = 5,
+}) => {
+  const searchElement = React.useRef<HTMLIonSearchbarElement>(null);
+
+  const [collapsed, setCollapsed] = React.useState(true);
+
+  const [showSearch, setShowSearch] = React.useState(false);
```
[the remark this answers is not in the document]
pkalita-lbl: Yeah that's better. I'm surprised my old Java habits of naming booleans with `is` or `has` didn't take over. Will update.

src/components/SampleList/SampleList.tsx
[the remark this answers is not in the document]
pkalita-lbl: Yeah `autofocus` was the first thing I thought of, but Ionic recommends against it: https://ionicframework.com/docs/developing/managing-focus. The `useEffect` approach is interesting. I'll give that a shot.

src/components/StudyList/StudyList.tsx
pkalita-lbl: Okay, maybe that's something we can get input from the wider group on.

src/components/SampleList/SampleList.tsx
pkalita-lbl: No it shouldn't be rendered in that case. Good catch.

src/components/SampleList/SampleList.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const [searchTerm, setSearchTerm] = React.useState<string | null>();
+
+  const samples = useMemo(
+    // This is reversed so that the most recent samples are shown first
```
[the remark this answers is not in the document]
pkalita-lbl: And thank you for your thoughtful reviews!

src/components/SampleList/SampleList.tsx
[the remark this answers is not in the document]
pkalita-lbl: When I made the change to swap `setTimeout` for `useEffect` some of the tests in `SampleList.test.tsx` started failing. Not because `useEffect` is a bad idea, but I think because some of my previous test writing had been a bit sloppy. So this really sent me down a bit of a rabbit hole (a productive one!) of figuring out better ways to structure component tests. It also led me to a eslint plugin that enforces better practices around using `testing-library`.

src/components/SampleList/SampleList.tsx
pkalita-lbl: I could go either way. I'll leave this one up to group feedback.

eecavanna (approved):
I looked at the changes that were introduced into this branch since the last time I reviewed it. Looks good to me! I'm comfortable with this being merged in.

PR #38, opened 2024-02-25
Implement tab-based navigation

eecavanna:
In this branch, I implemented tab-based navigation.

image

There are 3 tabs: 
- Studies
- Guide
- Settings

The `TabNavigator` component renders a horizontal bar containing those tabs; and also renders a router outlet for all routes that are "within" the tabbed section of the app. All such routes share a common route prefix (I arbitrarily set it to `/in`), which is matched by an application-level route that renders the `TabNavigator`, itself.

I also introduced `PlaceholderPage`, a temporary component we can use when we want an `IonPage` containing some basic content, in order to demonstrate functionality (e.g. a target to route to).

I also updated some relatively hard-coded paths I came across so that they use the common `PATHS` object instead. On a side note, I saw that there is a [draft PR](https://github.com/microbiomedata/nmdc-field-notes/pull/37/files#diff-9ff5b9d254197d871a6540cee2f27c8deb4e62c75b869e2cbf2fd265870ef9b7R20) in which the `PATHS` object will be replaced with something else—that's OK with me.

Finally, I swapped a couple routes back to an order that I, and I think one of my teammates, finds more intuitive (the previous order was the way it was because it was working, while the order we found more intuitive was not). It seems to me that the introduction of the tab navigation has changed how those two routes "behave" 🤷 :
```tsx
            <AuthRoute exact path={PATHS.STUDY_CREATE_PAGE}>
              <StudyCreatePage />
            </AuthRoute>
            <AuthRoute path={PATHS.STUDY_VIEW_PAGE}>
              <StudyViewPage />
            </AuthRoute>
```

### DIFF
--- a/src/Router.tsx
+++ b/src/Router.tsx
@@ -12,17 +12,22 @@ import { useStore } from "./Store";
 import LogoutPage from "./pages/LogoutPage/LogoutPage";
 import StudyViewPage from "./pages/StudyViewPage/StudyViewPage";
 import StudyCreatePage from "./pages/StudyCreatePage/StudyCreatePage";
+import TabNavigator from "./components/TabNavigator/TabNavigator";
+import PlaceholderPage from "./pages/PlaceholderPage";
 
 export const PATHS = {
   ROOT: "/",
-  HOME_PAGE: "/home",
+  HOME_PAGE: "/in/home",
   TUTORIAL_PAGE: "/tutorial",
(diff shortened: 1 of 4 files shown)


eecavanna:
I just made two more commits. Those were changes I had made before opening this PR, but forgot to commit.

eecavanna:
Thanks! I'll go ahead and merge this in.

Issue #28, opened 2024-02-09
Add deep linking into iOS and Android apps


pkalita-lbl:
With #26 the basic mechanisms for authentication are in place. I mainly tested using the web version locally (there is an issue with how the web version works on GitHub pages, see #27). The only missing piece for completing the login flow on iOS and Android versions is configuring them to handle links to `https://fieldnotes.microbiomedata.org` so that the app can collect the token provided by `nmdc-server`. Capacitor has a [handy guide](https://capacitorjs.com/docs/guides/deep-links) to follow.

eecavanna:
I'll start looking into the Apple/iOS part of this today.

eecavanna:
I immediately ran into a blocker: I don't have permission to register an App ID. I have contacted the "Account Holder" of our lab Apple Developer account to see if he can help.

In the meantime, here are some related links:
- Registering an app ID: https://developer.apple.com/help/account/manage-identifiers/register-an-app-id/
- App capabilities: https://developer.apple.com/documentation/xcode/adding-capabilities-to-your-app/
- "Identifiers" list: https://developer.apple.com/account/resources/identifiers/list

eecavanna:
Re-opening.

Current status:

- [x] Configure web server 
  - Done via PR #39
- [x] Configure Android app
  - Done via PR #40...
  - ...although I may have to provide someone with the Keystore file I generated
- [ ] Configure iOS app
  - TODO: https://capacitorjs.com/docs/guides/deep-links#add-associated-domain

PR #39, opened 2024-02-25
Create universal link-related files in `public` folder

eecavanna:
In this branch, I created two files:
- `apple-app-site-association`
- `assetlinks.json`

The first one will be used for Apple (iOS) universal links.

The second one will be used for Android universal links.

Both files are in `public/.well-known/`, so they will be included as-is in the build that gets deployed to GitHub Pages.

I also created a `README.md` file in the same directory. That file contains documentation about the directory and its contents.

---

While these changes tackle part of https://github.com/microbiomedata/nmdc-field-notes/issues/28, they do not tackle the entire ticket. Therefore, if GitHub automatically closes that ticket when this gets merged in, I will reopen the ticket.

### DIFF
--- a/public/.well-known/README.md
+++ b/public/.well-known/README.md
@@ -9,6 +9,7 @@ be included in the build (and, as a result, deployed to the web server) verbatim
 > like [this one](https://github.com/vitejs/vite/discussions/7374#discussioncomment-8557938) to delete it from the
 > build; but use a Node.js script instead of `rm` to delete the file,
 > since we know `node` will be present on the system (I think Windows offers `del` instead of `rm`).
+>
 > ```shell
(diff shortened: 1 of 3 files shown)


eecavanna:
I am merging this without review as I consider the changes to be harmless, it is currently not a workday, and I want to confirm these `.well-known` files will be accessible at the URLs I think they will.

I will use the following online tool to confirm the Android one is correct: https://developers.google.com/digital-asset-links/tools/generator

The Apple counterpart of that tool seems to no longer be online. There are several threads in the Apple Developer forum about it (for example, [this one](https://forums.developer.apple.com/forums/thread/741881)).

eecavanna:
🎉 The Android one is correct.

image

PR #40, opened 2024-02-25
Configure Android app to recognize incoming links

eecavanna:
In this branch, I updated the `AndroidManifest.xml` file—as instructed in the [Ionic docs](https://capacitorjs.com/docs/guides/deep-links#add-intent-filter) about universal links—so that the Android app would recognize incoming links.


eecavanna:
As with https://github.com/microbiomedata/nmdc-field-notes/pull/39, I will merge this in without review (for the same reasons).

PR #51, opened 2024-03-03
Decouple application code from environment variables

eecavanna:
In this branch, I introduced a layer of abstraction between the `import.meta.env` object and the rest of the application code. The new layer provides type information, documentation, and default values. It could also provide validation/sanitization/transformation in the future. I think the introduction of this new layer will make it easier for people to work on application code, as IDEs can now validate the (names and types of) ~environment variables~ "config variables" referenced by the application code.

Here's a pair of diagrams showing the "before" and "after" architecture (except that I accidentally depicted the new `config.ts` file as being outside of the application code, whereas it is part of the application code in reality).

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,11 @@ npm install
 
 ### Setup local environment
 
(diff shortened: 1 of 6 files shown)


pkalita-lbl (approved):
Makes sense to me

PR #60, opened 2024-03-13
Implement study edit and create screens

### LINKED ISSUE #45: Implement pages for updating existing study information and adding new studies
When tapping on the "Edit" link on the individual study view page (https://github.com/microbiomedata/nmdc-field-notes/issues/30), the user should go to a page that allows editing study-level information. When tapping on the "New" link from the list of studies on the home page (https://github.com/microbiomedata/nmdc-field-notes/issues/29), the user should go to a similar page that allows creating a new study. These pages should presumably share an underlying form component.

Mockup screenshots for reference
Image

Some thoughts:
* A "Delete Study" option was presented in the mockups for the study view page (https://github.com/microbiomedata/nmdc-field-notes/issues/30), but I think it's maybe more natural to present it in the edit page. I don't think it's something that will be used a lot, so I think it's okay to be an extra interaction away from the view page. The portal backend doesn't have an endpoint for deleting studies (or `SubmissionMetadata` objects, technically); I'll add another issue for implementing that.
* Validation rules for study-level information isn't defined in a formal schema; it's just part of the `nmdc-server` application code at the moment. I think we can just implement similar rules: study name, PI email, and template are required.
* The mockups show affordances for adding Contributors, but I think we can omit that for the initial version given that we're not focusing on multiple editors.

pkalita-lbl:
Fixes #45 

### Summary

These changes fill out the stub study create and study edit pages with real content. They both present a common study form component with basic validation. The edit page also includes a dropdown menu in the toolbar to initiate deleting a study.

### Details

* Add new dependency on [`react-hook-form`](https://react-hook-form.com/). I think most people have seen this library before (maybe?), so hopefully it doesn't need too much explanation. But this library provides form state management and validation. 
* Add methods to the `nmdc-server` client in `src/api.ts` for calling the `POST /api/metadata_submission` and `DELETE /api/metadata_submission/:id` endpoints. Since the `DELETE` endpoint returns a 204 response with an empty body I shuffled around the protected methods of `FetchClient` to avoid attempting to parse the body in that case.
* Add a new `useSubmissionCreate` hook which wraps `useMutation` and calls the `POST` endpoint method.
* Update the `useSubmission` hook to return a `deleteMutation` in addition to the existing `updateMutation` (which has been renamed from just `mutation`).
* Add a bunch more TypeScript interfaces to fully define the `MetadataSubmission` structure. Previously a lot of that structure was stubbed out with `object` types, and that was fine when we were only fetching data. We only needed to describe the parts that our application accessed. But now that we are sending data back to the server, we need to describe the whole structure. This is especially important because the backend application expects **everything** in the `metadata_submission` field to be filled out (even if it's just with empty strings or lists) when you create a new `SubmissionMetadata` record. Other top-level properties like `id` and `status` _don't_ need to be there, but the entire `metadata_submission` structure does. So that's also why there is a new `SubmissionMetadataCreate` interface that `SubmissionMetadata` now extends. `SubmissionMetadataCreate` describes what is necessary to create a `SubmissionMetadata` record and `SubmissionMetadata` describes the fully formed object you get back from the server.
* Add a new module (`src/data.ts`) with methods to generate a "blank" `SubmissionMetadataCreate` instance. 
* Add a new `StudyForm` component which displays the certain chosen fields within a `SubmissionMetadataCreate` instance and implements validation logic. The client of this component is responsible for providing the `SubmissionMetadataCreate` instance which provide the default form values and a callback that will be called when the user submits the form.
* Update the study edit page to fetch the submission, pass it to the the `StudyForm` component, and connect the submit callback to the appropriate mutation. This page also gets a dropdown menu in the toolbar for deleting a study.
* Update the study create page to generate a blank `SubmissionMetadataCreate` instance, pass it to the `StudyForm` component, and connect the submit callback to the appropriate mutation.

### DIFF
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -1,0 +1,211 @@
+import React from "react";
+import {
+  IonButton,
+  IonInput,
+  IonSelect,
+  IonSelectOption,
+  IonText,
+  IonTextarea,
+} from "@ionic/react";
+import SectionHeader from "../SectionHeader/SectionHeader";
+import { SubmissionMetadataCreate, TEMPLATES } from "../../api";
+import { Controller, useForm } from "react-hook-form";
+
+import styles from "./StudyForm.module.css";
+
+interface StudyFormProps {
+  submission: SubmissionMetadataCreate;
+  onSave: (submission: SubmissionMetadataCreate) => Promise<unknown>;
+}
+
+const EMAIL_REGEX =
+  /^(?!\.)(?!.*\.\.)([A-Z0-9_+-.]*)[A-Z0-9_+-]@([A-Z0-9][A-Z0-9-]*\.)+[A-Z]{2,}$/i;
+
+const StudyForm: React.FC<StudyFormProps> = ({ submission, onSave }) => {
+  const { handleSubmit, control, formState, setValue } =
+    useForm<SubmissionMetadataCreate>({
+      defaultValues: submission,
+      mode: "onTouched",
+      reValidateMode: "onChange",
+    });
+
+  return (
+    <form onSubmit={handleSubmit(onSave)}>
+      <div className="ion-padding">
+        <Controller
+          name="metadata_submission.studyForm.studyName"
+          rules={{ required: "This field is required" }}
+          control={control}
+          render={({ field, fieldState }) => {
+            return (
+              <IonInput
+                className={`${(fieldState.isTouched || formState.isSubmitted) && "ion-touched"} ${fieldState.invalid && "ion-invalid"}`}
+                labelPlacement="floating"
+                helperText="Provide a study name to associate with your samples."
+                errorText={fieldState.error?.message}
+                type="text"
+                onIonBlur={field.onBlur}
+                onIonInput={field.onChange}
+                {...field}
+              >
+                <div slot="label">
+                  Name<IonText color="danger">*</IonText>
+                </div>
+              </IonInput>
+            );
+          }}
+        />
+
+        <Controller
+          name="metadata_submission.studyForm.description"
+          control={control}
+          render={({ field, fieldState }) => (
+            <IonTextarea
+              className={`${(fieldState.isTouched || formState.isSubmitted) && "ion-touched"} ${fieldState.invalid && "ion-invalid"}`}
+              label="Description"
+              helperText="Provide a description of your study. This should include some general context of your research goals and study design."
+              labelPlacement="floating"
+              autoGrow
+              onIonBlur={field.onBlur}
+              onIonInput={field.onChange}
+              {...field}
+            />
+          )}
+        />
+
+        <Controller
+          name="metadata_submission.studyForm.notes"
+          control={control}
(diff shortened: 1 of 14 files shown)


src/components/StudyForm/StudyForm.tsx
yxu-lanl, line 1: The 'Save' button default is enabled. 
Screenshot 2024-03-13 at 4 53 06 PM

pkalita-lbl: Yeah that was an intentional choice. I can't say I'm an expert on the research, but I've heard UX people say that it's better to not disable buttons or show errors until after the user has interacted with the form (https://ux.stackexchange.com/a/76306, https://ux.stackexchange.com/a/9790, https://ux.stackexchange.com/a/17913).

yxu-lanl (approved):

eecavanna:
I'll review this now.

src/components/StudyForm/StudyForm.tsx
```diff
@@ -0,0 +1,211 @@
...
+        />
+
+        <Controller
+          name="metadata_submission.studyForm.piOrcid"
+          control={control}
+          render={({ field, fieldState }) => (
+            <IonInput
+              className={`${(fieldState.isTouched || formState.isSubmitted) && "ion-touched"} ${fieldState.invalid && "ion-invalid"}`}
+              label="ORCID iD"
+              helperText="ORCID iD of the Principal Investigator.."
```
eecavanna, line 149: Suggestion:

Three dots:

```diff
- helperText="ORCID iD of the Principal Investigator.."
+ helperText="ORCID iD of the Principal Investigator..."
```

Or no dots:

```diff
- helperText="ORCID iD of the Principal Investigator.."
+ helperText="ORCID iD of the Principal Investigator"
```

src/data.ts
```diff
@@ -0,0 +1,79 @@
...
+  Address,
+  AddressForm,
+  ContextForm,
+  MetadataSubmission,
+  MultiOmicsForm,
+  StudyForm,
+  SubmissionMetadataCreate,
+} from "./api";
+
+export const defaultAddress = (): Address => ({
```
eecavanna, line 11: I prefer function names be verbs (generally?). In this case, the verb that comes to mind, while attempting to keep things similarly concise, is `initAddress` (or `initializeAddress` or `resetAddress` or `makeDefaultAddress`). Same comment applies to all other functions in this file.

eecavanna (approved):
Looks great! I left some comments. I didn't come across any deal-breakers, with the caveat that I have not yet reviewed `src/queries.test.tsx` (I think it'll take me a long time, and I don't want to hold up the merge for it).

eecavanna:
I'm OK with this being merged in whenever @pkalita-lbl wants.

src/components/StudyForm/StudyForm.tsx
pkalita-lbl: It was supposed to be one period 😂. Bad copy-and-pasting from the submission portal.

eecavanna:
All the changes look good to me.

PR #72, opened 2024-03-21
Add ability to create new sample

### LINKED ISSUE #66: Add ability to create new sample
Currently tapping the "New" link from the Samples header on the Study page goes to a stub page. Instead it should:

* Insert an empty object into the submission's `sampleData`
* Once that modification is complete, navigate to the sample page for the newly created record

pkalita-lbl:
Fixes #66 

Previously the "New" link in the header of the Samples list on the Study view page was a router link to a stub sample creation page. Given that existing samples have a unified view/edit page, having a dedicated sample creation page doesn't really make a lot of sense. 

The behavior introduced with these changes is that when the "New" link is clicked, a new empty record is pushed into the submission's `sampleData` and the updated submission is sent to the server. Once that mutation completes successfully, the user is navigated to the view/edit page for the new sample. The sample creation stub page has been removed since it won't be used.


eecavanna (approved):
Looks good to me. (Reviewed via GitHub mobile app.)

yxu-lanl (approved):

PR #73, opened 2024-03-22
Add ability to delete samples

### LINKED ISSUE #67: Add ability to delete sample
Follow the same UI pattern that's implemented on the Study edit page:

Untitled

* Add an ellipsis button to the toolbar of the Sample page
* Tapping on the ellipsis menu presents a popup with a "Delete Sample" option
* Tapping on the "Delete Sample" option presents an alert to confirm.
* On confirmation, splice the sample out of the submission's `sampleData` and navigate to the Study page

pkalita-lbl:
Fixes #67 

Primarily this PR is about adding the ability to delete samples. This is implemented in `src/pages/SamplePage/SamplePage.tsx` similar to the way the delete study functionality works. However, in testing that functionality, I uncovered two bugs that I'm also addressing here.

First, I realized it was a mistake on my part to inject the `_index` field at the API layer. The only place it is really needed is in the `SampleList` component in order to keep sample indexes straight while filtering them. The downside of injecting the `_index` field way upstream in the API access methods is that it becomes part of the data that is sent back to the server when updating a sample. This, in an of itself, is not desirable since `_index` is a UI-only concern. But it also caused some confusion when looking up samples by `_index` to delete (which isn't even necessary -- `_index` is only necessary when filtering! but that's how I found this issue). So the fix here is to only inject those `_index` values at the layer it's needed: the `SampleList` component. 

Second, in testing I found a few validation-related bugs: 1) newly created samples weren't being validated and 2) issues when a previously valid sample gets partially re-validated as a field is updated. Those have also been fixed.

### DIFF
--- a/src/pages/SamplePage/SamplePage.tsx
+++ b/src/pages/SamplePage/SamplePage.tsx
@@ -93,9 +93,21 @@ const SamplePage: React.FC = () => {
           return;
         }
         if (result) {
-          draft[parseInt(sampleIndex)][modalSlot.name] = result;
+          // If the sample previously passed validation, its index won't be
+          // in `draft`. But it needs to be after this partial validation,
+          // so add a spot for it now.
+          if (!(sampleIndexInt in draft)) {
+            draft[sampleIndexInt] = {};
+          }
+          draft[sampleIndexInt][modalSlot.name] = result;
         } else {
-          delete draft[parseInt(sampleIndex)][modalSlot.name];
+          // If the sample previously passed validation, its index won't be
+          // in `draft`. Since it's still passing after this partial
+          // validation, just move on.
+          if (!(sampleIndexInt in draft)) {
+            return;
+          }
+          delete draft[sampleIndexInt][modalSlot.name];
         }
       }),
     );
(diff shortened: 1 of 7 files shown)


eecavanna:
I'll review this now.

src/pages/SamplePage/SamplePage.tsx
```diff
@@ -131,7 +202,7 @@ const SamplePage: React.FC = () => {
               slot={modalSlot}
               validationResult={
                 modalSlot
-                  ? validationResults?.[parseInt(sampleIndex)]?.[modalSlot.name]
+                  ? validationResults?.[sampleIndexInt]?.[modalSlot.name]
```
eecavanna, line 205: Note to self: 
- Here are the docs about "optional chaining":  
  - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

eecavanna (approved):

eecavanna:
I'm comfortable with this being merged in. 

Thanks for explaining the "why" underlying these changes, in the PR description.

I'm too far behind in my understanding of this part of the code base to make a specific comment about the changes. 😅

yxu-lanl (approved):

PR #75, opened 2024-03-22
Add option to getSubmissionSamples to allow creating missing sampleData fields when necessary

### LINKED ISSUE #74: Creating a new sample in a newly created study does not work
If I create a new study and then on the resulting sample view page create a new sample I wind up at `/in/study/c276b071-1fbf-4aaa-927a-7971c3bb63b0/sample/-1`. That `-1` at the end is wrong (should be `0`).

pkalita-lbl:
Fixes #74 

When a new submission is created it looks (in part) like:

```js
{
  ...
  metadata_submission: {
    ...
    sampleData: { }
  }
}
```

The existing logic of `getSubmissionSamples` would look for a field like `soil_data` in the `sampleData` object and if it wasn't there (as in the above case) then it would return a new empty array. This works fine for cases where the `sampleData` needs to be _displayed_. However when the `sampleData` is being _modified_ we need to ensure that the (for example) `soil_data` field is present in `sampleData` and set it to an empty array if it wasn't previously there.

The fix here to add an `options` parameter to `getSubmissionSamples` with a `createSampleDataFieldIfMissing` field. We set that field to `true` when creating new samples.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,7 +1,13 @@
 import { SubmissionMetadata, TEMPLATES } from "./api";
 import { SchemaDefinition, SlotDefinition } from "./linkml-metamodel";
 
-export function getSubmissionSamples(submission?: SubmissionMetadata) {
+export interface GetSubmissionSamplesOptions {
+  createSampleDataFieldIfMissing?: boolean;
+}
+export function getSubmissionSamples(
+  submission?: SubmissionMetadata,
+  options: GetSubmissionSamplesOptions = {},
+) {
   if (!submission) {
     return [];
   }
@@ -9,6 +15,12 @@ export function getSubmissionSamples(submission?: SubmissionMetadata) {
   const sampleDataField = TEMPLATES[environmentalPackageName]?.sampleDataSlot;
   if (!sampleDataField) {
     return [];
+  }
+  if (
+    !(sampleDataField in submission.metadata_submission.sampleData) &&
+    options.createSampleDataFieldIfMissing
+  ) {
+    submission.metadata_submission.sampleData[sampleDataField] = [];
   }
   return submission.metadata_submission.sampleData[sampleDataField] || [];
 }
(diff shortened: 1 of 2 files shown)


yxu-lanl (approved):

eecavanna:
I'll review this now.

src/utils.ts
```diff
@@ -10,6 +16,12 @@ export function getSubmissionSamples(submission?: SubmissionMetadata) {
   if (!sampleDataField) {
     return [];
   }
+  if (
+    !(sampleDataField in submission.metadata_submission.sampleData) &&
+    options.createSampleDataFieldIfMissing
+  ) {
+    submission.metadata_submission.sampleData[sampleDataField] = [];
```
eecavanna, line 23: The only thing I don't understand about this change is why the default value is an array (and not something that depends upon the _type_ associated with whatever `submission.metadata_submission.sampleData[sampleDataField]` is. I think this is just a matter of me not knowing about the latter variable. I'll chalk this up to that—and I'll make myself a note to learn about the latter (for one thing, so I can more effectively review code).

eecavanna (approved):
Looks good to me. (I have some "homework" to do.)

src/utils.ts
pkalita-lbl: It's a fair question. It might be helpful to think about the structure of `sampleData` in terms of the desktop submission portal's interface. 

image

The keys of `sampleData` correspond to the tabs (e.g. the "Soil" tab corresponds to the `soil_data` key in `sampleData`). Within the "Soil" tab/`soil_data` key the array elements correspond to rows in the spreadsheet -- what we refer to as individual "samples". So the underlying data corresponding to that screenshot looks something like:

```js
{
  metadata_submission: {
    sampleData: {
      soil_data: [
        { samp_name: "samp 1", ... },  // <- this is a sample
        { samp_name: "samp 2", ... },  // <- this is another sample
        { samp_name: "samp 3", ... }, 
        { samp_name: "samp 4", ... }, 
      ],
      emsl_data: [ 
        ... // this is stuff we'll never show in the app
      ],
      jgi_mg_data: [ 
        ... // this is stuff we'll never show in the app
      ],
      jgi_mt_data: [ 
        ... // this is stuff we'll never show in the app
      ],
    }
  }
}
```

eecavanna: Ah, OK, thanks! It makes sense to me now. Thanks for providing the visual aids.

PR #100, opened 2024-04-24
Fix vertical alignment of version number on "Settings" screen (on Android/MD)

eecavanna:
In this branch, I fixed the vertical alignment of the version number on the "Settings" screen (on Android/MD devices). It was already correct on iOS devices.

Before:

image

After:

image

I also updated `README.md` so that it explicitly lists the query string developers can use to force the styles to be Android/MD, in an attempt to facilitate testing with both styles in the future.


eecavanna:
I am ready for this to be merged in.

pkalita-lbl (approved):

PR #104, opened 2024-04-26
Fix scrolling issue in sample edit modal

### LINKED ISSUE #84: Unable to save changes of sample metadata when save button out of frame
Note: this is for after GSP/ ECR :-)

Just something I noticed on my phone, but I can replicate if I reduce the window size on my browser:
 - Open a sample and click on a metadata to modify, one with a long documentation text (e.g. broad-scale environmental context)
 - If the window is too small, the "Cancel" and "Save" buttons are not visible. 
 - To make them visible, one can grad the frame up, but then the frame goes back down as soon as it's released, so I can't click on "save"

Possible solutions:
- Make the frame "stick" when pulling it up
- Put the "Cancel" and "Save" button above the doc ?

Screenshot below is to illustrate what happens when the window is too small (cancel and save out of frame)

image

pkalita-lbl:
Fixes #84 

These changes allow the sample edit modal contents to be scrolled. This means that on a small-ish device, when the contents of the modal are long (e.g. in when editing the "broad-scale environmental context" slot), you will be able to scroll down to see the save button. 

Relevant technical links in code comments.

### DIFF
--- a/src/components/SampleSlotEditModal/SampleSlotEditModal.module.css
+++ b/src/components/SampleSlotEditModal/SampleSlotEditModal.module.css
@@ -1,3 +1,21 @@
+ion-modal.sampleSlotEditModal {
+  /**
+   * This is the height of the modal when it is in the "expanded" state. The docs recommend this
+   * approach over having a max breakpoint < 1. See:
+   *   - https://ionicframework.com/docs/api/modal#custom-sheet-height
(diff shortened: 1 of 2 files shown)


eecavanna (approved):
Thanks for including those comments in the CSS (giving background info about the values).

eecavanna:
LGTM and I'm comfortable with this being merged in.

yxu-lanl (approved):

PR #107, opened 2024-04-27
Allow users to toggle dark mode via the "Settings" screen

eecavanna:
In this branch, I added a "Color scheme" selector to the "Settings" page.

#### Terminology

Before I continue, here are three terms related to this branch:
1. Color scheme preference: This is the `(prefers-color-scheme: dark)` thing that can be read via a CSS media query. It is a user's general color scheme preference, not specific to this app.
2. Color palette mode: This is the thing a user can select on the "Settings" screen. It is either `Light`, `Dark`, or `System` (explained below).
3. Color palette: This is the set of CSS styles that ultimately gets applied to the app's UI. It is either a light color palette (the default) or a dark color palette.

#### Behavior

Here's how the color palette mode selector (which is labeled "Color scheme" in the UI) works.

When the user selects the "Light" color palette mode, that selection is persisted to the store and the **light** color palette is applied to the app's UI. Similarly, when the user selects the "Dark" color palette mode, that selection is persisted to the store and the **dark** color palette is applied to the app's UI.

When the user selects the "System" color palette mode, that selection is persisted to the store. Also, the app checks whether the user generally prefers dark color schemes. If so, the **dark** color palette is applied to the app's UI—otherwise, the **light** color palette is applied to the app's UI.

Here's a flowchart showing the same thing:

```mermaid
graph LR
  U[User selects<br>color palette mode] --> Mode
  Mode{Color palette mode}
  Mode --> |Light| CPL[Color palette = Light]
  Mode --> |Dark| CPD[Color palette = Dark]
  Mode --> |System| PCS{General color<br>scheme preference}
  PCS -->|dark| CPD
  PCS -->|"(anything else)"| CPL
```

#### Notes

This was my first time adding a new property to the store in this project.


src/theme/colorScheme.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Returns `true` if the value is a valid color scheme.
+ */
+export const isValidColorScheme = (value: unknown) => {
+  return Object.values(ColorScheme).includes(value as ColorScheme);
+};
+
+/**
+ * Toggles the dark color scheme according to the flag passed in.
+ *
+ * Note: Activate/deactivating the dark color scheme is a matter of adding/removing
```
eecavanna, line 17:
```suggestion
 * Note: Activating/deactivating the dark color scheme is a matter of adding/removing
```

src/components/SystemColorSchemePreferenceMonitor/SystemColorSchemePreferenceMonitor.tsx
```diff
@@ -0,0 +1,42 @@
...
+  useLayoutEffect(() => {
+    const onChangePreference = (event: MediaQueryListEvent) => {
+      console.debug("System prefers dark color scheme: ", event.matches);
+      if (colorScheme === ColorScheme.System) {
+        toggleDarkColorScheme(event.matches);
+      }
+    };
+
+    // Set up an event listener that listens for changes in the system's color scheme preference.
+    // Reference: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList
```
eecavanna, line 30:
```suggestion
    // Add the event listener.
```

src/theme/colorScheme.ts
```diff
@@ -0,0 +1,60 @@
...
+export const toggleDarkColorScheme = (wantsDarkColorScheme: boolean) => {
+  const htmlEl = document.documentElement;
+  htmlEl.classList.toggle("ion-palette-dark", wantsDarkColorScheme);
+};
+
+/**
+ * Checks whether the system prefers the dark color scheme.
+ */
+export const doesSystemPreferDarkColorScheme = () => {
+  const prefersDark = window.matchMedia("(prefers-color-scheme: dark)").matches;
```
eecavanna, line 33: Note: `src/components/SystemColorSchemePreferenceMonitor/SystemColorSchemePreferenceMonitor.tsx` also creates a `mediaQueryList` with the same parameter. Opportunity to consolidate?

src/theme/colorScheme.ts
```diff
@@ -0,0 +1,60 @@
...
+  return prefersDark;
+};
+
+/**
+ * Activates the specified color scheme.
```
eecavanna, line 38: Consider distinguishing between the set of things the user selects from (i.e. "Light", "Dark", and "System") and the set of palettes that can be applied to the app (i.e. "Light" and "Dark"). These terms may be useful: `mode`, `palette`, `scheme`.

src/theme/colorScheme.ts
eecavanna: Consolidated.

src/theme/colorScheme.ts
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Note: Activating/deactivating the dark color scheme is a matter of adding/removing
+ *       the `.ion-palette-dark` class from the `<html>` element.
+ *
+ * Reference: https://ionicframework.com/docs/theming/dark-mode#css-class
+ *
+ * @param wantsDarkColorScheme {boolean} Whether you want the dark color scheme to be enabled.
+ */
+export const toggleDarkColorScheme = (wantsDarkColorScheme: boolean) => {
+  const htmlEl = document.documentElement;
```
eecavanna, line 37: Selecting this on each call is unnecessary (although I don't think there is any significant performance penalty). Consider defining this one time, near the top of this module, outside of any functions.

yxu-lanl (approved):

pkalita-lbl (approved):
Looks great! This turned out to be a bit more involved than I thought it would be; thanks for sticking with it. I guess I naively thought that Ionic provided something out-of-the-box like what you implemented in `src/theme/colorPalette.ts`. Live and learn!

PR #116, opened 2024-06-03
Install `madge` NPM package and eliminate circular dependencies

### LINKED ISSUE #115: Fix and prevent further circular imports
In working on #92 I realized that some of the changes I was making had inadvertently created a bunch of circular imports. JavaScript is generally tolerant of circular imports, and sometimes (often?) they don't cause undesired behavior. However there's no guarantee of that. I think even benign circular imports somewhat lower the maintainability of a codebase. 

In the branch where I'm working on #92 I used [`dpdm`](https://github.com/acrazing/dpdm) to find circular imports. It turned out that some of the ones I found were preexisting. So I plan on splitting out the work of fixing the ones that already existing in `main` today and also add a check to our PR workflow to prevent new ones from entering.

pkalita-lbl:
Fixes #115 

### Summary

These changes add a dependency on the [`madge`](https://github.com/pahen/madge) package and connect it to a new `check.imports` script in `package.json`. The script is now called in the tests workflow to ensure we don't introduce new circular dependencies. In #115 I talked about using [`dpdm`](https://github.com/acrazing/dpdm), but I realized after making the issues that `madge` seems to be more popular and has more people working on it. 

As far as fixing the existing circular imports, all of them looked something like this:

```mermaid
flowchart LR
    A(Router.tsx)
    A --> B(various pages & components)
    B --> A
```

Where the various pages & components were importing `Router.tsx` to get the `paths` object.

So the fix is to move the `paths` object to a separate module and you end up with:

```mermaid
flowchart TD
    A(paths.ts)
    B(Router.tsx) --> A
    C(various pages & components) --> A
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,9 +41,11 @@ jobs:
       - name: Install npm packages
         run: npm ci
       - name: Check formatting
-        run: npm run format:check
+        run: npm run check.format
       - name: Lint code
         run: npm run lint
+      - name: Check for circular imports
+        run: npm run check.imports
       - name: Run build
         run: npm run build
       - name: Run unit tests
--- a/README.md
+++ b/README.md
@@ -103,7 +103,6 @@ npm run lint
 > That tells npm you want it to run the script named `lint`, defined in the `package.json` file.
 > At the time of this writing, that script runs `eslint src`.
 
-
 ### Run circular import check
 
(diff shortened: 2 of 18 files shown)


eecavanna (approved):
I'm comfortable with this being merged in.

In contrast to Python, I haven't given much thought to circular imports in JavaScript.

yxu-lanl (approved):

Issue #121, opened 2024-06-10
App uses generic Ionic favicon (instead of Field Notes-specific favicon)


eecavanna:
The [favicon in use](https://github.com/microbiomedata/nmdc-field-notes/blob/main/public/favicon.png) is still the generic/default/boilerplate Ionic favicon; instead of a Field Notes-specific favicon (e.g. the NMDC logo).

Issue #120, opened 2024-06-10
PWA is not installable and does not set theme color for address bar


eecavanna:
The PWA report generated by Lighthouse (in Chrome DevTools—screenshot below) lists these shortcomings:
- Web app manifest or service worker do not meet the installability requirements
  > Service worker is the technology that enables your app to use many Progressive Web App features, such as offline, add to homescreen, and push notifications. With proper service worker and manifest implementations, browsers can proactively prompt users to add your app to their homescreen, which can lead to higher engagement. [Learn more about manifest installability requirements](https://developer.chrome.com/docs/lighthouse/pwa/installable-manifest/?utm_source=lighthouse&utm_medium=devtools).
  >
  > ##### Failure reason
  > No supplied icon is at least 144 px square in PNG, SVG or WebP format, with the purpose attribute unset or set to "any"
- Does not set a theme color for the address bar. Failures: 
    - ```html
      No `<meta name="theme-color">` tag found.`
      ```
      > The browser address bar can be themed to match your site. [Learn more about theming the address bar](https://developer.chrome.com/docs/lighthouse/pwa/themed-omnibox/?utm_source=lighthouse&utm_medium=devtools).

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/134325062/3196f9fe-5e6c-4263-a8fe-d96150472ad8)

eecavanna:
The `manifest.json` file contains references to `webp` images at `../icons/*.webp`. I don't know what that refers to. If the web app is deployed to `example.com`, the `manifest.json` file will be at `example.com/manifest.json`. In that case, `../` doesn't make sense to me. 🤷

eecavanna:
The `manifest.json` file also uses the MIME type of `image/png` for images of file type `webp`, which I don't think is correct.

```js
    {
      "src": "../icons/icon-48.webp",
      "type": "image/png",
      "sizes": "48x48",
      "purpose": "any maskable"
    },
```

Why was that done that way?

PR #122, opened 2024-06-10
Make PWA installable and customize favicon

eecavanna:
In this branch, I fixed all PWA-related issues reported by Lighthouse. Those were:
> 1. Web app manifest or service worker do not meet the installability requirements. [...] No supplied icon is at least 144 px square in PNG, SVG or WebP format, with the purpose attribute unset or set to "any"
> 2. No `<meta name="theme-color">` tag found.

While fixing the first one, I noticed the `manifest.json` file had some bugs; specifically, it referred to images that didn't exist, and the image file suffixes did not agree with the specified MIME types. I fixed those things in this PR also.

I used the open source image editor, "gimp", to create the favicon.

- Fixes https://github.com/microbiomedata/nmdc-field-notes/issues/120
- Fixes https://github.com/microbiomedata/nmdc-field-notes/issues/121


eecavanna:
I'm ready for this PR to be reviewed/merged.

pkalita-lbl (approved):
I tested this by running an iPhone 15 simulator, visiting my local dev server on Safari, and using the "Add to Home Screen" feature. The resulting shortcut from the home screen seems to work fine and looks mostly app-ish. The main difference is the blue toolbar:

image

Versus in the true app:

image

That may be a fundamental limitation of the PWA architecture. Regardless, I'm not expecting too many people to use this as a PWA so I'm not worried about it.

eecavanna:
Thanks for testing it. I set the "theme-color" (HTML `meta` element) to match our Ionic theme's primary color. I think that is what caused the header bar to be that blue color. We could use a less prominent color from our color palette instead. I do think some "theme-color" is required, though, based on the report from Lighthouse.

pkalita-lbl:
I think what you chose is the best option given the constraints of PWAs

Issue #93, opened 2024-03-27
Router gets in a funky state after deleting sample


pkalita-lbl:
Steps to reproduce:

1. From home page, navigate to a study with samples
2. From study page, navigate to an individual sample
3. From the sample page, use the ellipsis menu in the header to delete it. If successful, this should result in the user being navigated back to the study page.
4. From the study page, use the back button in the header to return to the home page

At this point, the router is a bit confused about which page should be displayed as evidenced by a ghostly version of the study page header displayed under the home page header:

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/102547565/b31736e9-0059-4541-a27b-9366ad58e4e2)

Behind the scenes, Ionic keeps multiple pages in the DOM so that it can do the animated transitions between them. For example, at the end of step 2 above the relevant portion of the DOM look like this:

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/102547565/b31a47b8-e697-4da8-af90-d779d5df843b)

You can see a stack of three `div` elements with the `ion-page` class. The bottom two are hidden, and the top two have the `can-go-back` class. I believe that means that tapping the "Back" button at that point would pop the top one off the stack and unhide the middle (or what will become the top) one.

This all seems to get jumbled up when we use the `router.push()` function provided by `useIonRouter`. This is the relevant portion of the DOM after the completion of step 4 above:
 
![image](https://github.com/microbiomedata/nmdc-field-notes/assets/102547565/0342f64e-6625-46e0-83dc-8460a9798afa)

We have two `div` elements with the `ion-page` class and **both** are visible.

I think the warrants a [deeper dive into Ionic routing](https://ionicframework.com/docs/react/navigation), but the solution might be to -- instead of always using `router.push()` after deleting an item -- use a combination of `router.canGoBack`, `router.goBack`, and `router.push` to **prefer** navigating backwards after the delete if possible and if not fallback to pushing a new route.

pkalita-lbl:
After taking a second look at this, I think what's going on in Eric's use case is a little more understandable. 

[...]

PR #123, opened 2024-06-10
Show logged in user information in Settings, provide a link for logging out

### LINKED ISSUE #78: Add User section to Settings page
On the Settings tab, add a User section. It should display the name of the currently logged in user and a button to logout. We have a dedicated logout page, but I'm not sure we really need it. Consider whether it makes sense for the logout button of the Settings page to call `setApiToken` and redirect. If so we could probably remove the dedicated logout page.

pkalita-lbl:
Fixes #78 

### Summary

The main purpose of these changes is to add the ability to view the currently logged-in user and provide a link to log out from the Settings tab. These changes actually highlighted some routing issues. So there is an attempt to clean that up in here as well. I believe these changes will help with #93, but I'll do more focused testing of that issue after these changes are in and close it if there is nothing else to do.

### Details

* In `src/Store.tsx` the `StoreProvider` now fetches user details from the `/api/me` endpoint whenever the logged-in state changes. It saves the response in its internal state and provides it through the `StoreContext` value. A `null` value is provided when no user is logged in.
* The `src/pages/SettingsPage/SettingsPage.tsx` component has been visually reorganized into seconds (User, Appearance, and About). Each of the individual sections is implemented as separate, new component (e.g. `src/components/SettingsUserList/SettingsUserList.tsx`). I did a little refactoring of the display of the existing settings to follow the "stacked label" format we use throughout the app. I believe these sections will also be appropriate for new settings that I know we need to add.
* The addition of the log out button in Settings which attempts a redirect back to the welcome page initially caused two problems. 
  * Because we were reading the `isLoggedIn` state in `src/Router.tsx` to change the redirect of the root path, the entire `Router` component was getting rerendered on logout and this ended up causing an infinite loop of reredners. I solved that by moving that redirect logic to a separate component (`src/pages/RootPage/RootPage.tsx`).
  * Next the redirect to the welcome path wasn't working because the `TabNavigator` had a fallback route that (for reasons still slightly unclear to me) was getting confused try to the redirect _out_ of the tabs. That led me to eventually conclude that we should only have _one_ fallback at in `Router.tsx` and that we should also limit the amount of navigating we do within tabs by pulling all the Study and Sample related routes up to the root level. 
* Remove the extraneous `/login` and `/logout` routes and their associated page components. We weren't really using those anywhere.
* Allow logged out users to view the tabs route, but show a fallback message in the `HomePage` component for logged out users. In the future we can explore if we can show studies from the offline cache in this case instead.
* I attempted to standardize our nomenclature around using "log in" an "log out" as verbs.
* In reviewing other apps it seemed like "Theme" was a standard way to label dark vs light mode and that "System default" was often use to label the "do whatever my OS is doing" option.


yxu-lanl (approved):

src/pages/HomePage/HomePage.tsx
```diff
@@ -15,7 +18,14 @@ const HomePage: React.FC = () => {
...
       </IonHeader>
       <IonContent>
-        <StudyList />
+        {isLoggedIn ? (
+          <StudyList />
+        ) : (
+          <div slot="fixed">
+            You must <a onClick={initiateLogin}>log in</a> to view and edit
+            studies
+          </div>
```
eecavanna, line 27: I think I'd prefer this without the "You must":
```diff
- You must <a onClick={initiateLogin}>log in</a> to view and edit
- studies
+ <a onClick={initiateLogin}>Log in</a> to view and edit studies
```

I don't consider this to be a deal-breaker, though.

src/paths.ts
```diff
@@ -1,9 +1,9 @@
 const IN = "/in";
-const STUDY = `${IN}/study`;
+const STUDY = `/study`;
 
 const paths = {
   root: "/",
-  home: STUDY,
+  home: `${IN}${STUDY}`,
```
eecavanna, line 6: Note to self (@eecavanna) to look into this on my own: this path has the prefix, but paths like `studyCreate` don't.

eecavanna (approved):
Looks good to me! Thanks for implementing this. I'm comfortable with it being merged in as is.

PR #125, opened 2024-06-12
Update sample search index on changes

### LINKED ISSUE #91: New sample is not searchable immediately after creation
I made a study with 5 samples, named `SAMP-001` through `SAMP-005`. When I view that study and use the searchbar to filter the sample list using the string "003" I can find the `SAMP-003` sample. If I create a new sample named `SAMP-006` and then go back to the study view page, when I search for "006", the new sample is not found.

It would appear that the index built by `react-minisearch` is not getting updated at the right time.

pkalita-lbl:
~~Leaving this in draft because I don't want to merge it and update the web version until after Emiley is back from ASM Microbe.~~

---

Fixes #91 

### Summary

The `useMiniSearch` hook we use to set up a [`MiniSearch`](https://github.com/lucaong/minisearch) search index in the `SampleList` component accepts a documents array, but it **does not** track changes to this list ([by design](https://github.com/lucaong/react-minisearch/issues/43#issuecomment-2109657251)). This means that as the samples list changes (e.g. a new one is added) the search index doesn't follow along. These changes use a `useEffect` callback to perform a re-indexing whenever the samples list changes.

Separately I noticed as I was updating the tests that we were getting a bunch of spurious warnings ([for example](https://github.com/microbiomedata/nmdc-field-notes/actions/runs/9456057271/job/26047184327#step:10:23)). This was resolved by simply updating to the latest versions of `@testing-library/react` and `@testing-library/user-event`.


eecavanna (approved):
Looks good to me. Thanks for linking to the "by design" conversation, and for documenting your rationale for using "remove all, add all" in `useEffect`. I'm comfortable with this being merged in.

yxu-lanl (approved):

PR #126, opened 2024-07-02
Add ability to control which slots are displayed on sample view/edit page

### LINKED ISSUE #80: Add ability to control which slots are displayed on sample view/edit page
Currently when a user goes to the sample view/edit page, they see all the slots for appropriate schema class. This is often _a lot_ of slots. Some of them will probably never be collected in the field and others will not apply to certain studies/scenarios. 

Proposal: Use local app storage to store, for each schema class, which slots should be displayed. That is, the locally stored data would look something like:

```js
// Store the slot names that the user wants to see, keyed by the class name
{
  AirInterface: ["samp_name", "alt", "humidity"],
  SoilInterface: ["samp_name", "depth", "elev"],
}
```

From a user perspective if they get to a sample page showing a class that is _not_ in the settings object (e.g. `WaterInterface` following the above example), show a banner at the top that says something like "Too many fields? Tap here to select fields to show". Tapping that message should bring up a modal overlay which lists all the slots with checkboxes and a Select/Deselect All control. The modal should explain that once set, only the selected fields will be shown for that template. And that they can change their selections in Settings later. The next time the user lands on a sample page where the template _is_ in the settings object, show a message at the bottom of the list indicating that some fields are hidden based on previous selection and that they can change it in Settings.

In the Settings tab, add a section for Field Visibility. In that section, show each template that is represented in the stored settings object (e.g. Air and Soil following the above example). Tapping on a template name should bring up the modal with the checkboxes where they can update their selections.

pkalita-lbl:
Fixes #80 

### Summary

These changes give the user the ability to show and hide slots (described as "fields" in the interface) in the sample view/edit interface for each schema class (described as "templates" in the interface). Showing and hiding slots does _not_ have any effect on what it sent to the backend or saved in the database. It is strictly a filter at the view layer.

The first time a user views a sample for, say, the soil template, they will see a message asking if the want to customize the slot list. If they choose to customize they will see a modal overlay where each slot is presented along with its description, an indication of whether it is required by the schema, and a checkbox to show or hide it. Once the list is customized, a message at the bottom of the sample view/edit interface is shown which provides access to the modal again.

| Banner | Modal | Footer |
| --- | --- | --- |
| image | image | image |

The same controls are also available via the Settings tab. There is a new Field Visibility entry in the Appearance section. Tapping this item leads to a new page where each relevant schema class is listed along with an indication of whether its list of slots has already been customized. Tapping on an item in this list brings up the same modal as before.

| Settings Page | Field Visibility Page | 
| --- | --- |
| image | image |

### Details

* The `StoreProvider` now provides two new functions, `getHiddenSlotsForSchemaClass` and `setHiddenSlotsForSchemaClass`. These functions access and modify the new internal state, `hiddenSlots`, which is simply a mapping between class names and lists of hidden slot names. The setter function also saves the `hiddenSlots` object to persistent storage. When the `StoreProvider` is initialized, it checks for a value persistent storage.
* The new `SlotSelector` component is a stateless, purely presentational component responsible for rendering a list of slot groups with checkboxes and executing a callback with the correct list of selected slots as the checkboxes are toggled. The checkboxes attached to group labels can go into an indeterminate state if some slots within that group are checked and some aren't. The component also allows specifying a list of slots names which should be rendered as a disabled checkbox. See tests and stories for more details.
* The new `SlotSelectorModal` component is a stateful component responsible for accepting a template name, looking up which slots should be hidden for that template according to the store, presenting a `SlotSelector` in a modal, and saving any changes back to storage when the modal closes.
* The new `FieldVisibilitySettingsPage` component is responsible for rendering a list of templates and opening a `SlotSelectorModal` when one is selected. It is added to the Router at the `/settings/field-visibility` path.
* The existing `SampleView` has been updated to display messages to the user based on whether the sample's template has or doesn't have hidden slots. It also now has a `SlotSelectorModal` which is opened by these messages.
* Vite has built-in support for [PostCSS](https://vitejs.dev/guide/features.html#postcss). I've added a PostCSS config file along with the [PostCSS Nesting](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-nesting) plugin as a new dev dependency. This allows Sass-_like_ rule nesting (see `src/components/SampleView/SampleView.module.css`) but using a standards-track CSS proposal instead of Sass's bespoke syntax. This is inline with Vite's own [recommendation](https://vitejs.dev/guide/features.html#css-pre-processors) regarding CSS preprocessors. 
* The new `RequiredMark` component is my submission for this year's Nobel Prize in Computer Science.


yxu-lanl (approved):

src/components/RequiredMark/RequiredMark.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { IonText } from "@ionic/react";
+
+const RequiredMark: React.FC = () => {
```
eecavanna, line 4: shocked-magikarp-omg

src/components/SampleView/SampleView.tsx
```diff
@@ -18,47 +22,100 @@ function formatSlotValue(value: SampleDataValue) {
...
+      {hiddenSlots !== undefined && hiddenSlots.length > 0 && (
+        <IonList className="ion-padding-bottom">
+          <IonItem lines="none" onClick={() => setIsModalOpen(true)}>
+            <IonLabel class="ion-text-wrap">
+              <p>
+                Not seeing a field you were looking for? Tap here to update
+                field visibility settings.
+              </p>
+            </IonLabel>
+          </IonItem>
```
eecavanna, line 110: I am wondering whether any accessibility issues (e.g. inability to do "keyboard navigation" on the PWA, or some iOS accessibility helper not treating it like an interactive element) will come up due to this `IonItem` not having an `href` or `button` property set. I see a section in the Ionic docs about "Clickable Items": https://ionicframework.com/docs/api/item#clickable-items

src/components/SlotSelector/SlotSelector.tsx
```diff
@@ -0,0 +1,135 @@
...
+            if (index >= 0 && !disabled) {
+              draft.splice(index, 1);
+            }
+          });
+        }
+      }),
+    );
+  };
+
+  const groupState = (slotGroup: SlotGroup) => {
```
eecavanna, line 71: Thanks for including those comments in this function's body!

src/components/SlotSelectorModal/SlotSelectorModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <IonButton onClick={handleSave} strong={true}>
+              Save
+            </IonButton>
+          </IonButtons>
+        </IonToolbar>
+      </IonHeader>
+      <IonContent>
+        <div className="ion-padding">
+          <p>
+            Select the fields you would like to see by default when entering
```
eecavanna, line 106: I'm a little confused about the word "default" here. Maybe it's there to tell the user that it will be the new normal, but they can always change it later. Assuming I understand the user-facing behavior, I think I'd find this more clear:

```diff
- Select the fields you would like to see by default when entering
+ Select the fields you would like to see when entering
```

src/pages/FieldVisibilitySettingsPage/FieldVisibilitySettingsPage.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <ThemedToolbar>
+          <IonButtons slot="start">
+            <IonBackButton defaultHref={paths.settings} />
+          </IonButtons>
+          <IonTitle>Field Visibility</IonTitle>
+        </ThemedToolbar>
+      </IonHeader>
+      <IonContent>
+        <p className="ion-padding">
+          Select a template to customize which fields are visible by default.
```
eecavanna, line 61: This is another place where I am a little confused about the phrase "by default".

Assuming I understand the user-facing behavior, I think I'd find something like this to be more clear:

```diff
- Select a template to customize which fields are visible by default.
+ Select a template to customize which fields will be visible when using that template.
```

eecavanna (approved):
Looks great! Thanks for including that PR description, which set the stage for the diff (in my experience reviewing the PR).

I left a few comments. I didn't see anything I consider to be a deal breaker.

src/components/RequiredMark/RequiredMark.tsx
eecavanna: I will nominate you for a Nobel Prize in aster`*`...ophysics!

eecavanna:
I'm comfortable with this branch being merged in as is.

src/Store.tsx
```diff
@@ -151,16 +177,53 @@ const StoreProvider: React.FC<PropsWithChildren> = ({ children }) => {
     }
   }
 
+  /**
+   * Returns the hidden slots for the specified schema class.
+   */
+  function getHiddenSlotsForSchemaClass(className?: string) {
+    if (className === undefined) {
+      return undefined;
```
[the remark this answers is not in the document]
pkalita-lbl: I can add this as a comment, but the assumption is that `undefined` means the user has never explicitly made any choice about the fields for the given schema class and `[]` means they have explicitly said "I want to see them all".

src/components/SampleView/SampleView.tsx
pkalita-lbl: Good thought. I will take a closer look at that.

src/components/SlotSelectorModal/SlotSelectorModal.tsx
pkalita-lbl: I think I was trying to communicate that the preference is "sticky" across all existing and new samples for that template -- as opposed to a sample-by-sample thing. But, yeah, looking back at this it's a little confusing. I'll simplify it.

src/components/SampleView/SampleView.tsx
pkalita-lbl: I added the `button` and `detail={false}` props to `IonItem` as suggested by the docs.

src/Store.tsx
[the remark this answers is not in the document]
pkalita-lbl: I updated the signature of `getHiddenSlotsForSchemaClass` to not accept `undefined` and added a comment clarifying what different return values mean.

src/components/SampleView/SampleView.tsx
eecavanna: Cool—thank you!

Issue #124, opened 2024-06-11
Format of app name below app icon is too dense (iOS)


eecavanna:
Screenshot:

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/134325062/ad5f4fa6-3ace-4c79-b628-b8653d955100)

pkalita-lbl:
As far as I can tell we are setting the [right key](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundledisplayname) with the string `NMDC Field Notes`. It's not clear to me why the spaces are getting squeezed out. Is that an iOS thing perhaps?

pkalita-lbl:
Oh, perhaps we need to remove the `CFBundleName` [key](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundlename) since we do want a name longer than 15 characters?

eecavanna:
Next steps:
- [ ] Find out what our options are
- [ ] Consult the rest of the squad; in particular, the engagement team members
- [ ] Implement what we decide

eecavanna:
When I set `CFBundleDisplayName` to `Field Notes`, here's what the app icon looks like:

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/134325062/216ede08-c6c5-41bb-bc0c-a7c9315d6644)

eecavanna:
When I set `CFBundleDisplayName` to `NMDC Field Notes`, here's what the app icon looks like:

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/134325062/ebe538e9-a45c-4186-9b0f-02aab8958215)

eecavanna:
When I set `CFBundleDisplayName` to `NMDC`, here's what the app icon looks like:

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/134325062/45660f2d-023f-4207-9953-c9b0d405adb4)

eecavanna:
When I set `CFBundleDisplayName` to `NMDC FieldNotes` (which is exactly 15 characters), here's what the app icon looks like:

![image](https://github.com/microbiomedata/nmdc-field-notes/assets/134325062/de5b6049-f6fb-47e3-b164-2225b1b19ffb)

> There are no ellipses shown.

PR #127, opened 2024-07-08
Shorten iOS app icon text to "Field Notes" to avoid ellipses

eecavanna:
In this branch, I configured the iOS app so that the text below the icon on the iOS home screen  says "Field Notes" (the app icon already says "nmdc" on it) instead of "NMDCFieldNot...".

iOS seems to me to impose a 15-character limit on this string. Optionally, see the comments in #124 for details.

Before:

image

After:

image


pkalita-lbl (approved):
Seems reasonable to me, esp. considering that the icon has NMDC in it

PR #130, opened 2024-07-15
Allow user to populate Study PI Name and ORCID iD with user's own info

eecavanna:
In this branch, I added buttons on the right sides of two text inputs on the "Create Study" screen:
- Principal Investigator > Name
- Principal Investigator > ORCID iD

The buttons are only rendered when the user is logged in (so, if we end up allowing logged-out users to snoop around the UI, these buttons will not appear to them).

When the user clicks a button, the app will update the field so it contains the associated value from the user's own ORCID account. For example, if the user clicks the button on the "Name" field, the app will update the "Name" field so it contains the user's name.

When the field already contains the value that clicking the button would cause the field to have, the button is disabled.

Here's a brief screencast of the buttons in action (with personal information blurred here):

https://github.com/user-attachments/assets/d118223a-5360-4064-8e0c-683fc750e02a

Here are some things I think reviewers may request changes about:
1. Which icon I used (currently a wand icon)
2. Which color I used (currently the primary color from the theme)
3. The tooltip text that appears when the user hovers over the buttons (currently "Use my name" and "Use my ORCID iD")

### DIFF
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -133,7 +133,7 @@ const StudyForm: React.FC<StudyFormProps> = ({ submission, onSave }) => {
                   slot={"end"}
                   title={"Use my name"}
                   aria-label={"Use my name"}
-                  onClick={() => field.onChange(loggedInUser.name)}
+                  onClick={() => setValue(field.name, loggedInUser.name)}
                   disabled={field.value === loggedInUser.name}
                 >
                   <IonIcon
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -134,6 +134,7 @@ const StudyForm: React.FC<StudyFormProps> = ({ submission, onSave }) => {
                   title={"Use my name"}
                   aria-label={"Use my name"}
                   onClick={() => field.onChange(loggedInUser.name)}
+                  disabled={field.value === loggedInUser.name}
                 >
                   <IonIcon
                     slot={"icon-only"}
@@ -198,6 +199,7 @@ const StudyForm: React.FC<StudyFormProps> = ({ submission, onSave }) => {
                   title={"Use my ORCID iD"}
                   aria-label={"Use my ORCID iD"}
                   onClick={() => field.onChange(loggedInUser.orcid)}
+                  disabled={field.value === loggedInUser.orcid}
                 >
                   <IonIcon
                     slot={"icon-only"}
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -133,11 +133,7 @@ const StudyForm: React.FC<StudyFormProps> = ({ submission, onSave }) => {
                   slot={"end"}
                   title={"Use my name"}
                   aria-label={"Use my name"}
-                  onClick={() => {
-                    // TODO: Populate the field with the logged-in user's name.
-                    const { name } = loggedInUser;
-                    alert(`TODO: Populate field with ${name}`);
-                  }}
+                  onClick={() => field.onChange(loggedInUser.name)}
                 >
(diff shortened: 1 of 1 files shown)


yxu-lanl (approved):

src/components/StudyForm/StudyForm.tsx
```diff
@@ -120,7 +125,26 @@ const StudyForm: React.FC<StudyFormProps> = ({ submission, onSave }) => {
...
-            />
+            >
+              {/* If the user is logged in, show a button they can press to use their name. */}
+              {loggedInUser !== null ? (
+                <IonButton
+                  fill={"clear"}
+                  slot={"end"}
+                  title={"Use my name"}
+                  aria-label={"Use my name"}
+                  onClick={() => field.onChange(loggedInUser.name)}
```
pkalita-lbl, line 136: I think invoking `onChange` manually is discouraged (https://react-hook-form.com/docs/usecontroller/controller). I would use the `setValue` function which is already being unpacked from the `useForm` call.

```suggestion
                  onClick={() => setValue(field.name, loggedInUser.name)}
```

src/components/StudyForm/StudyForm.tsx
```diff
@@ -166,7 +190,26 @@ const StudyForm: React.FC<StudyFormProps> = ({ submission, onSave }) => {
...
-            />
+            >
+              {/* If the user is logged in, show a button they can press to use their ORCID iD. */}
+              {loggedInUser !== null ? (
+                <IonButton
+                  fill={"clear"}
+                  slot={"end"}
+                  title={"Use my ORCID iD"}
+                  aria-label={"Use my ORCID iD"}
+                  onClick={() => field.onChange(loggedInUser.orcid)}
```
pkalita-lbl, line 201: Ditto above.

```suggestion
                  onClick={() => setValue(field.name, loggedInUser.orcid)}
```

pkalita-lbl (commented):
I like the functionality and the UI. Just a couple suggestions about the implementation.

src/components/StudyForm/StudyForm.tsx
eecavanna: Thanks, @pkalita-lbl. I was confused about that. I interpreted the `Controller` documentation (which I think has a grammatical error in it) as saying:

> "...you should avoid manually invoking `setValue`..."

image

That led to me searching for alternatives. I saw, in other places on the (documentation) page, they kept using `field.onChange`, so that's why I went with that one.

I'll switch to `setValue`. Maybe the author of the documentation meant to write:

```diff
This prop update `formState` and you should avoid 
- manually invoke 
+ manually invoking it and, instead, invoke
`setValue` or other API related to field update.
```

eecavanna:
I'm going to commit the suggested changes, but I want to test it locally before it gets merged in. I'll report back after I've tested it locally.

eecavanna:
I confirmed it worked locally with the suggested changes in place.

I'm wondering whether the current behavior will be too "surprising" to some users and whether they'd prefer that the button display a modal that explains what the autofill would do, and the user can then confirm or cancel.

I'll go ahead and merge this in as is, given the review feedback. We can go back and add the modal later if necessary.

src/components/StudyForm/StudyForm.tsx
pkalita-lbl: Yeah I'm nearly certain that's what the documentation is trying to say. The `react-hook-form` documentation unfortunately has a lot of such typos.

pkalita-lbl:
> I'm wondering whether the current behavior will be too "surprising" to some users and whether they'd prefer that the button display a modal that explains what the autofill would do, and the user can then confirm or cancel.

I think that, even if you didn't know ahead of time what the button is going to do, the result is quite understandable. And if you don't like it, the recourse to undo it is so clear (just edit the field manually) that we don't need to go out of our way to warn people about it.

PR #134, opened 2024-07-19
Set source_client to field_notes when generating new submission

### LINKED ISSUE #128: Add "from Field Notes" flag to `SubmissionMetadata` records
Currently there is no way to differentiate a `SubmissionMetadata` record that was created via the Field Notes app versus the submission portal. If nothing else we should at least keep track of which studies were created via the app for metrics purposes.

The tasks here are:

1. Add a way to keep track of the source of a submission to the `nmdc-server` Postgres database. This could be as simple as a boolean column like `is_from_field_notes` on the `submission_metadata`. Or, for future expansibility, we might want a text column like `source` that can take an enumerated set of values (including "submission portal" and "field notes", for example).
2. In `nmdc-field-notes`, set this value appropriately when making a new submission.

pkalita-lbl:
Fixes #128 

This is a follow-on to https://github.com/microbiomedata/nmdc-server/pull/1300.

When a new submission is generated by the Field Notes app (`src/data.ts`), set the `source_client` field to `field_notes`. 

I made one of the mock submissions used in tests (`src/mocks/fixtures.ts`) have a `source_client` of `submission_portal` just to ensure that we don't do anything with the `source_client` field that implicitly assumes a value of `field_notes` coming _from_ the backend.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -169,6 +169,7 @@ export interface User {
 
 export interface SubmissionMetadataCreate {
   metadata_submission: MetadataSubmission;
+  source_client: "submission_portal" | "field_notes" | null;
 }
 
 export interface SubmissionMetadata extends SubmissionMetadataCreate {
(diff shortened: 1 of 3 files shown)


eecavanna (approved):
Looks good to me! 

FYI @yxu-lanl, I found this straightforward to review after having reviewed the corresponding changes in `nmdc-server`. Basically, when either the submission portal or the mobile app creates a submission, it will include this information saying what type of client initialized the submission.

PR #137, opened 2024-07-24
Obtain editing lock on submissions

### LINKED ISSUE #90: App should respect the edit lock set by the backend
When a user edits a submission via the desktop submission portal, a few lock fields get set/cleared on the `submission_metadata` record, for example:

```js
{
  ...
  lock_updated: "2024-03-21T23:07:09.397242"
  locked_by: {
    id: "2f836e03-c2ac-4494-a641-dae1b77f7ece"
    orcid: "0000-0002-6150-307X"'
    name: "Patrick Kalita"
    is_admin" true
  }
}
```

The app should respect the locks and not allow editing of a study that's already being edited elsewhere. Additionally it should attempt to claim a lock when a user starts editing a study.

pkalita-lbl:
Fixes #90 

### Summary

`SubmissionMetadata` records have two fields (`locked_by` and `lock_updated`) that indicate who is currently editing a submission. Previously we were not consulting this field at all which could lead to edit conflicts if more than one person were to edit a submission at the same time. So the changes here introduce three things:

1. Obtaining a lock prior to making submission modifications (this includes creating a new sample, editing an existing study or sample, or deleting an existing study or sample)
2. If obtaining the lock fails (typically because someone else has it), display an informative message to the user
3. If obtaining the lock succeeds, release the lock once the editing is complete. This is generally done when the user navigates away from an editing interface. There's no guarantee this navigation will actually happen (i.e. a user starts editing a sample and then drops their phone in a river). The desktop submission portal faces a similar issue which is why the backend already automatically expires these locks after 30 minutes.

### Details

* The `nmdc-server` backend provides endpoints to request and release a submission lock. These endpoints are called by new methods in `src/api.ts` along with some refinement to the type definitions. 
* In `src/queries.ts` two new [mutations](https://tanstack.com/query/latest/docs/framework/react/guides/mutations) have been added to the call the API methods and to keep the server data state updated when the requests settle. 
* In the study edit page (`src/pages/StudyEditPage/StudyEditPage.tsx`), we attempt to obtain the submission lock when the page loads. If the lock cannot be obtained, the entire study form and the delete study button become disabled, and we present a message to the user indicating who holds the lock. If obtaining the lock is successful it is released when leaving the page.
* In the sample page (`src/components/SampleView/SampleView.tsx`), we similarly attempt to obtain the submission lock when the page loads. If the lock cannot be obtained, the slot editing modal and the delete sample button become disabled, and we present a message to the user indicating who holds the lock. If obtaining the lock is successful it is released when leaving the page.
* In the study view page (`src/components/StudyView/StudyView.tsx`), we do _not_ obtain a lock when simply viewing the page. We _do_ attempt to obtain the lock immediately before attempting to create a new sample in the study. If obtaining the lock fails, a new sample is not created and a message is displayed to the user.
* One slightly tangential change here is in `src/queries.ts`. One feature of the submission portal is that when you create or update a submission and you provide an ORCID iD for the study's PI, that user should also have permission to view/edit the submission. For whatever reason that isn't handled automatically by the backend. Instead it's up to the client to use the `permissions` field in the request body to set that. 

### Testing

If you want to test out these changes you need to have a submission that is locked by someone other than yourself. And it also helps if you are an admin so that you can access everyone's submissions. I find it easiest to do that by direct SQL commands to my local `nmdc-server` Postgres database. For example, check to see if your are an admin:

```sql
select * from user_logins where name = '<your name>';
```

If `is_admin` is false:

```sql
update user_logins set is_admin = true where name = '<your name>';
```

Get the 10 most recently created submissions:

```sql
select * from submission_metadata order by created desc limit 10;
```

Pick one of those submissions and a random person's `id` from the `user_logins` table and set the lock:

```sql
update submission_metadata
set locked_by_id = '<other user id>', lock_updated = now()
where id = '<recent submission id>';
```

As mentioned previously a lock that is older than 30 minutes is ignored by the backend, so if you want to check which submissions still have an active lock:

```sql
select * from submission_metadata
inner join user_logins on submission_metadata.locked_by_id = user_logins.id
where lock_updated + interval '30 minutes' >= now()
order by lock_updated desc;
```

Once you have a submission with a fresh lock from someone else you can try various operations in the interface to make sure you're not allowed to modify the submission and that you're notified that someone else holds the lock.


src/api.ts
```diff
@@ -201,7 +211,14 @@ export interface TokenResponse {
   expires: number;
 }
 
-class ApiError extends Error {
+export interface LockOperationResult {
+  success: boolean;
+  message: string;
+  locked_by?: User | null;
+  lock_updated?: string | null;
```
eecavanna, line 218: I recommend adding a comment saying what the string is (assuming it's not arbitrary); e.g. `// timestamp` or `// ISO 8601-formatted timestamp`

src/api.ts
```diff
@@ -348,6 +365,24 @@ class NmdcServerClient extends FetchClient {
     });
   }
 
+  async getSubmissionLock(id: string) {
```
eecavanna, line 368: This is one situation where I think `get...` is more ambiguous than I would like. I've gotten used to `get...` being like "read"—as opposed to "claim". I'm wondering whether using a synonym/alternative here would help highlight that we are doing something other than _reading_ here. Ideas:
- `claim`
- `acquire` (synonym of `get`, but breaks me free of assumptions about `get`)
- `obtain` (same as above)

To clarify, I don't consider `get...` to be a dealbreaker.

src/components/StudyView/StudyView.tsx
```diff
@@ -128,6 +139,11 @@ const StudyView: React.FC<StudyViewProps> = ({ submissionId }) => {
           <SampleList
             submission={submission.data}
             onSampleCreate={handleSampleCreate}
+            sampleCreateFailureMessage={
+              lockMutation.isError
+                ? `Cannot create new sample because this study is currently being edited by ${submission.data.locked_by?.name || "unknown user"}`
```
eecavanna, line 144: Suggestion (not requirement). I'll defer to others for the final decision here (i.e. keep it how it is or add "an").

```suggestion
                ? `Cannot create new sample because this study is currently being edited by ${submission.data.locked_by?.name || "an unknown user"}`
```

src/pages/SamplePage/SamplePage.tsx
```diff
@@ -186,12 +212,27 @@ const SamplePage: React.FC = () => {
...
               </IonItem>
             </IonList>
           </IonContent>
         </IonPopover>
+
+        {!loggedInUserCanEdit && (
+          <Banner color="warning">
+            <IonLabel>
+              Editing is disabled because this sample is currently being edited
+              by {submission.data?.locked_by?.name || "unknown user"}
```
eecavanna, line 231:
```suggestion
              by {submission.data?.locked_by?.name || "an unknown user"}
```

src/pages/StudyEditPage/StudyEditPage.tsx
```diff
@@ -103,24 +126,45 @@ const StudyEditPage: React.FC = () => {
...
               </IonItem>
             </IonList>
           </IonContent>
         </IonPopover>
 
+        {!lockMutation.isPending && !loggedInUserCanEdit && (
+          <Banner color="warning">
+            <IonLabel>
+              Editing is disabled because this study is currently being edited
+              by {submission.data?.locked_by?.name || "unknown user"}
```
eecavanna, line 145:
```suggestion
              by {submission.data?.locked_by?.name || "an unknown user"}
```

src/queries.test.tsx
```diff
@@ -13,8 +13,14 @@ import {
   useSubmissionList,
 } from "./queries";
 import { produce } from "immer";
-import { patchMetadataSubmissionError, server, delay } from "./mocks/server";
+import {
+  patchMetadataSubmissionError,
+  server,
+  delay,
+  acquireLockConflict,
```
eecavanna, line 20: I see you have used the word `acquire` here. I added a comment to a line way above this one (in a different file) about using a word other than `get` in a function name. In that context—if the options were `acquire` and `get`—I'd prefer `acquire`.

src/queries.test.tsx
```diff
@@ -274,7 +281,82 @@ test("useSubmission should delete submission", async () => {
...
+
+  // Verify that the mutation completes successfully and that the lock information is updated
+  await waitFor(() =>
+    expect(result.current.unlockMutation.isSuccess).toBe(true),
+  );
+  expect(result.current.query.data?.locked_by).toBeNull();
+});
+
 test("useSubmissionCreate should create submission", async () => {
+  // When SubmissionMetadata objects are returned they do not contain a full list of people who
```
eecavanna, line 342:
```suggestion
  // When SubmissionMetadata objects are returned, they do not contain a full list of people who
```

src/queries.test.tsx
```diff
@@ -274,7 +281,82 @@ test("useSubmission should delete submission", async () => {
...
+  // have permission roles on the submission. So to verify that the PI (not the same as the logged
```
eecavanna, line 343:
```suggestion
  // have permission roles on the submission. So, to verify that the PI (not the same as the logged
```

eecavanna (approved):
Thanks for including that PR description! 👏 

I left some comments/suggestions on the diff. I didn't encounter anything in the diff that I considered to be a deal-breaker, so I am comfortable with this branch being merged in as is (I'll be OOO from Fri-Tue).

I have a question regarding the 30-minute lock timeout. Suppose Alice obtains the lock on a Study at 1 PM and she is still actively editing the Study at 1:25 PM. Does her lock still expire at 1:30 PM, or does that 30-minute expiration keep getting pushed back each time she makes a change to the Study?

src/api.ts
pkalita-lbl: I think that's a good idea. Especially because it's literally not a `GET` request. I think in the comments I already mostly used the word "acquire".

Issue #135, opened 2024-07-20
Update iOS app assets to reflect new Apple Developer Account team


eecavanna:
On July 19, 2024, we were given access to a different Lab-related team on the Apple Developer website. This team has a distinct "Team ID" from the one we were using before. The task here is to identify the various parts of our code base and external things we've set up (e.g. certificates in Xcode) to work with this new team.

Dependents: Completion of https://github.com/microbiomedata/nmdc-field-notes/issues/114 depends upon completion of this issue.

eecavanna:
| Original team ID | New team ID | 
| -- | -- |
| `4L4Y2A43SA` | `6G3V3EPHUC` |

The team ID appears in:
- [public/.well-known/apple-app-site-association](https://github.com/microbiomedata/nmdc-field-notes/blob/d6f04529b2303a27bdab58e626d9a2fce9a49bde/public/.well-known/apple-app-site-association#L6)
- [ios/App/App.xcodeproj/project.pbxproj](https://github.com/microbiomedata/nmdc-field-notes/blob/d6f04529b2303a27bdab58e626d9a2fce9a49bde/ios/App/App.xcodeproj/project.pbxproj#L353)

eecavanna:
While the originally-registered app identifier still existed in the "ENT" account, I tried—in the non-"ENT" account—creating an app identifier having the same Bundle ID. The Apple Developer website prevented me from doing so. Here's the error message it showed. I am not surprised by this, but did want to confirm it.

![image](https://github.com/user-attachments/assets/50f0c202-56b9-45ab-88a9-b5643102e13a)

eecavanna:
This is done, and deep linking still works.

PR #138, opened 2024-07-25
Replace Apple Developer "ENT" Team ID with non-"ENT" Team ID

eecavanna:
Today, I deleted the NMDC Field Notes app identifier from the "ENT" Apple Developer account and created an NMDC Field Notes app identifier in the non-"ENT" Apple Developer account. The reason I first deleted it from the "ENT" account is so that I could use the same Bundle ID in the non-"ENT" account. The actions I took are documented in [this Issue comment](https://github.com/microbiomedata/nmdc-field-notes/issues/135#issuecomment-2249553410).

In this PR branch, I have updated all references to the Team ID so that, instead of referring to the Team corresponding to the "ENT" account (where our app identifier no longer exists), they refer to the Team corresponding to the non-"ENT" account (where our app identifier now exists).

Since I'll be out Friday, and given the specific changes involved, I may end up merging this in tonight—without review—so I can proceed to test (and, if needed, repair) deep link functionality sooner than I would otherwise be able to.

### DIFF
--- a/ios/App/App.xcodeproj/project.pbxproj
+++ b/ios/App/App.xcodeproj/project.pbxproj
@@ -350,7 +350,7 @@
 				CODE_SIGN_ENTITLEMENTS = App/App.entitlements;
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
-				DEVELOPMENT_TEAM = 4L4Y2A43SA;
+				DEVELOPMENT_TEAM = 6G3V3EPHUC;
 				INFOPLIST_FILE = App/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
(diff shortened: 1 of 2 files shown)


eecavanna:
I am merging this in without review. Rationale:
- It's after work hours
- I want to test their effects on deep link functionality (e.g. ORCID login in the iOS simulator)
- Tomorrow is the last day of the sprint where I'll be in the office
- I am confident the changes won't impact the PWA or Android versions of the app
- I'm the only person I'm aware of that uses the iOS version of the app
- The changes are straightforward for me to revert if needed

eecavanna:
After making this change, I confirmed that deep linking (both ORCID login and tapping on `fieldnotes.microbiomedata.org/*` links in other apps) still worked.

For reference: It didn't work at first/immediately. In that situation, I used Xcode to "clean the build folder" and then I used `npm run sim.ios` to retry simulating the app. At this point, deep linking worked.

Issue #42, opened 2024-02-29
Project `README.md` doesn't say an API server is necessary


eecavanna:
While running the app locally (PWA served by Vite via `$ npm run dev`), I visited the `/in/study` page, which normally shows a list of studies. In this case, it said "No studies yet". 

It took me a couple minutes to realize my local `nmdc-server` environment was not running (I think I spun it down within the past few days). That is user error on my part, but I can see myself making the same mistake in the future.

I'm creating this ticket to represent the task of either:
- Displaying something on the app UI that indicates whether the app can access its API server or not. Since we're already planning for there to be "offline" functionality, so maybe this ticket is effectively a nudge to raise the priority of an aspect of that—some kind of status indicator aspect.
- Adding a note to the `README.md` about ensuring an API server is up and running.
- Both of the above.

eecavanna:
I created this ticket while facing a CORS error, but I think this applies to any situation where the app has Internet access, but can't access the API server.

PR #146, opened 2024-08-01
Fix permissions update

pkalita-lbl:
This fixes a bug that Montana just reported to me that I had already fixed in a development branch for #42. That branch isn't quite ready for a PR, so I'm just cherry picking a couple of commits out of it.

The issue was introduced in #137, and the problem has to do with attempting to mutate an existing object in order to add the `permissions` field instead of using `produce` to perform the mutation.

Since this is a fairly egregious bug that prevents updating submissions we should produce and distribute new iOS and Android builds after this is merged in.

### DIFF
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -41,18 +41,22 @@ export const submissionKeys = {
 
 export function addDefaultMutationFns(queryClient: QueryClient) {
   queryClient.setMutationDefaults(submissionKeys.details(), {
-    mutationFn: async (updated: SubmissionMetadata) => {
+    mutationFn: async (updated: SubmissionMetadataUpdate) => {
       await queryClient.cancelQueries({
         queryKey: submissionKeys.detail(updated.id),
       });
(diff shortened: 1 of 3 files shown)


src/queries.ts
```diff
@@ -41,18 +41,22 @@ export const submissionKeys = {
 
 export function addDefaultMutationFns(queryClient: QueryClient) {
   queryClient.setMutationDefaults(submissionKeys.details(), {
-    mutationFn: async (updated: SubmissionMetadata) => {
+    mutationFn: async (updated: SubmissionMetadataUpdate) => {
```
eecavanna, line 44: Side note (**not** a deal breaker for this PR): I think I'd prefer a different variable name from `updated`; e.g. `updatedSubmissionMetadata` or `updatedSM` or something. This is not the focus of this PR/these changes—I just wanted to express that thought.

eecavanna (approved):

PR #157, opened 2024-08-16
Improved offline support and API error indicators

### LINKED ISSUE #42: Project `README.md` doesn't say an API server is necessary

pkalita-lbl:
Fixes #42 

### Summary

This is a bit of a pick-n-mix of change all in the neighborhood of API error messaging _and_ offline support, since those two things are closely related. The expected functionality can be boiled down to:

* While offline:
  * Do not allow requests to create or delete a submission. This might be erring on the defensive side, but I think allowing those would lead to more weird edge cases than we're prepared to handle right now.
  * Do allow requests to update a submission. Such requests will be paused until the client is back online. Successive requests to update to the same submission should replace earlier requests so that one cumulative update request per submission goes out later.
  * Skip the requests to lock and unlock submissions. While offline we know that such a request can't be fulfilled, so in order to allow updates we simply have to _presume_ that the lock is available.
* While online
  * Let all requests go out across the network. 
  * If any network request (query or mutation) is in progress, show a top-level loading indicator
  * If a network request fails (because the server is unreachable _or_ the server responded with a non-2xx status), show an error indicator to the user
  * A failed query should _not_ prevent cached data from being shown

### Details

* The new `NetworkStatusProvider` provides a single `isOnline` boolean value which can be accessed through the `useNetworkStatus` hook. Under the hood this provider uses the `@capacitor/network` plugin to get the initial online status and setup a lister to respond to changes in online status. React Query has its own [OnlineManager](https://tanstack.com/query/latest/docs/reference/onlineManager) which is how it knows when to pause or retry requests. However, I found there were certain cases where React Query wasn't as reliable as Capacitor at detecting connection changes. So the `NetworkStatusProvider` also attempts to keep them in sync by manually setting React Query's network status.
* The initialization of `StoreProvider` has been updated to account for cases where the app starts up while offline. Specifically: 
  * When a user logs in, store _all_ of their user details in persistent storage (in addition to their refresh token, which was all that was previously stored)
  * When the app starts up, if user details are found in persistent storage use them to set the logged-in user state.
  * When the app starts up, if a refresh token is found in persistent storage immediately provide it to the `nmdc-server` API client. Previously this happened only _after_ a successful token exchange took place (which obviously can't happen if the app starts while offline).
  * When the app starts up _and_ we're online, attempt the token exchange and fetch updated user details from the server. If the app starts up while offline, this token exchange won't happen. Instead, when the client does go online, the first network request will go out and fail with an `Unauthorized` response, the client will detect that, perform the token exchange at that point, and then retry the request. That logic was in place before, but the API client didn't have the refresh token it needed (see previous bullet point).
* Instances of `IonProgressBar` have been removed individual components (e.g. `StudyList`). Instead, one _global_ `IonProgressBar` has been added to `ThemedToolbar`. It responds to _any_ query or mutation managed by React Query. `ThemedToolbar` has also been updated to show a banner when the user is offline.
* Two new components, `QueryErrorBanner` and `MutationErrorBanner`, are responsible for displaying error messages when network requests fail. Both of those components are thin wrappers around the new `ErrorBanner` component, which is responsible for actually formatting the error message. This component does show (although not by default) the nitty-gritty technical details of the failed request. I have mixed feelings about this, but I have a feeling it might be useful at least during the beta testing period.

| image | image |
| --- | --- |

* I added a [pull-to-refresh](https://ionicframework.com/docs/api/refresher) capability to the study list and individual study pages. Even though React Query is pretty smart about refetching at the right time, I think it's just comforting to be able to refresh on demand.
* Lots of little logic changes around when certain requests go out or certain operations are allowed based on network status.

### Testing

* Offline support is probably best tested using an Android emulator. Airplane mode on the device can be used to force the app into an offline state. I would recommend pointing to the dev `nmdc-server` instance, though, because even in airplane mode the device knows it can make local (e.g. `127.0.0.1`) requests. A web browser's dev tools can also simulate offline conditions but you _cannot_ test starting the app in an offline state that way.
* To test the API error messages (while online) I used my local `nmdc-server` instance and have it unconditionally raise exceptions in specific request handler functions (i.e. in `nmdc_server/api.py`). Sometimes I would also intentionally add delays in order to verify the transitions between loading and error states. So for example:
```python
@router.get(
    "/metadata_submission",
    tags=["metadata_submission"],
    responses=login_required_responses,
    response_model=query.MetadataSubmissionResponse,
)
async def list_submissions(
    db: Session = Depends(get_db),
    user: models.User = Depends(get_current_user),
    pagination: Pagination = Depends(),
):
    query = crud.get_submissions_for_user(db, user)
    time.sleep(2)
    raise HTTPException(status_code=410, detail="whoops")
    return pagination.response(query)
```

### DIFF
--- a/src/NetworkStatus.tsx
+++ b/src/NetworkStatus.tsx
@@ -1,0 +1,62 @@
+import React, {
+  createContext,
+  PropsWithChildren,
+  useContext,
+  useEffect,
+  useState,
+} from "react";
+import { Network } from "@capacitor/network";
+import { PluginListenerHandle } from "@capacitor/core";
+
+interface NetworkStatusContextValue {
+  isOnline: boolean;
+}
+
+const NetworkStatusContext = createContext<NetworkStatusContextValue>({
+  isOnline: true,
+});
+
+const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
+  const [isOnline, setIsOnline] = useState(true);
+
+  useEffect(() => {
+    // Used to prevent setting up the listener from being added from an unmounted component
+    let ignore = false;
+    let listener: PluginListenerHandle | null = null;
+
+    async function addNetworkListener() {
+      const currentStatus = await Network.getStatus();
+      // If the component is unmounted before the previous await finishes, ignore the result and
+      // no not set up a listener.
+      if (ignore) {
+        return;
+      }
+      setIsOnline(currentStatus.connected);
+
+      listener = await Network.addListener("networkStatusChange", (status) => {
+        setIsOnline(status.connected);
+      });
+    }
+
+    void addNetworkListener();
+
+    return () => {
+      if (listener) {
+        void listener.remove();
+      }
+      ignore = true;
+    };
+  }, []);
+
+  return (
+    <NetworkStatusContext.Provider value={{ isOnline }}>
+      {children}
+    </NetworkStatusContext.Provider>
+  );
+};
+
+export default NetworkStatusProvider;
+
+export const useNetworkStatus = () => {
+  return useContext(NetworkStatusContext);
+};
--- a/src/NetworkStatus.tsx
+++ b/src/NetworkStatus.tsx
@@ -7,6 +7,7 @@ import React, {
 } from "react";
 import { Network } from "@capacitor/network";
 import { PluginListenerHandle } from "@capacitor/core";
+import { onlineManager } from "@tanstack/react-query";
 
 interface NetworkStatusContextValue {
   isOnline: boolean;
@@ -31,9 +32,11 @@ const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
       if (ignore) {
         return;
       }
+      onlineManager.setOnline(currentStatus.connected);
       setIsOnline(currentStatus.connected);
 
       listener = await Network.addListener("networkStatusChange", (status) => {
+        onlineManager.setOnline(currentStatus.connected);
         setIsOnline(status.connected);
       });
     }
@@ -48,6 +51,7 @@ const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
     };
   }, []);
 
+  // Don't render the children until we know for certain the initial online status
   if (isOnline === null) {
     return null;
   }
--- a/src/NetworkStatus.tsx
+++ b/src/NetworkStatus.tsx
@@ -26,17 +26,25 @@ const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
     let listener: PluginListenerHandle | null = null;
 
     async function addNetworkListener() {
-      const currentStatus = await Network.getStatus();
+      const initialStatus = await Network.getStatus();
       // If the component is unmounted before the previous await finishes, ignore the result and
-      // no not set up a listener.
+      // do not set up a listener.
       if (ignore) {
         return;
       }
-      onlineManager.setOnline(currentStatus.connected);
-      setIsOnline(currentStatus.connected);
+
+      // React Query does **not** get notified of the initial online status. It assumes an active
+      // network connection by default. Since the app could be opened in an online state, we need to
+      // inform React Query of the initial online status.
+      // See: https://tanstack.com/query/latest/docs/reference/onlineManager
+      onlineManager.setOnline(initialStatus.connected);
+      setIsOnline(initialStatus.connected);
 
       listener = await Network.addListener("networkStatusChange", (status) => {
-        onlineManager.setOnline(currentStatus.connected);
+        // When the network status changes, React Query **should** be capable of tracking the
+        // changes, but there shouldn't be any harm in manually updating the status here, just in
+        // case. Then we know that Capacitor and React Query are in sync.
+        onlineManager.setOnline(status.connected);
         setIsOnline(status.connected);
       });
     }
@@ -44,9 +52,12 @@ const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
     void addNetworkListener();
 
     return () => {
+      // If the component is unmounted after the listener is set up, remove it.
       if (listener) {
         void listener.remove();
       }
+
+      // Always flag the component as unmounted when the cleanup function is called.
       ignore = true;
     };
   }, []);
--- a/src/NetworkStatus.tsx
+++ b/src/NetworkStatus.tsx
@@ -17,7 +17,7 @@ const NetworkStatusContext = createContext<NetworkStatusContextValue>({
 });
 
 const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
-  const [isOnline, setIsOnline] = useState(true);
+  const [isOnline, setIsOnline] = useState<boolean | null>(null);
 
   useEffect(() => {
     // Used to prevent setting up the listener from being added from an unmounted component
@@ -47,6 +47,10 @@ const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
       ignore = true;
     };
   }, []);
+
+  if (isOnline === null) {
+    return null;
+  }
 
   return (
     <NetworkStatusContext.Provider value={{ isOnline }}>
--- a/src/NetworkStatus.tsx
+++ b/src/NetworkStatus.tsx
@@ -22,14 +22,14 @@ const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
 
   useEffect(() => {
     // Used to prevent setting up the listener from being added from an unmounted component
-    let ignore = false;
+    let isComponentUnmounted = false;
     let listener: PluginListenerHandle | null = null;
 
     async function addNetworkListener() {
       const initialStatus = await Network.getStatus();
       // If the component is unmounted before the previous await finishes, ignore the result and
       // do not set up a listener.
-      if (ignore) {
+      if (isComponentUnmounted) {
         return;
       }
 
@@ -58,7 +58,7 @@ const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
       }
 
       // Always flag the component as unmounted when the cleanup function is called.
-      ignore = true;
+      isComponentUnmounted = true;
     };
   }, []);
 
--- a/src/QueryClientProvider.tsx
+++ b/src/QueryClientProvider.tsx
@@ -9,11 +9,13 @@ import { useStore } from "./Store";
 import { createAsyncStoragePersister } from "@tanstack/query-async-storage-persister";
 import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 
+const GARBAGE_COLLECTION_TIME = 1000 * 60 * 60 * 24 * 7; // 1 week
+
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       staleTime: 1000 * 20, // 20 seconds
-      gcTime: 1000 * 60 * 60 * 24 * 7, // 1 week
+      gcTime: GARBAGE_COLLECTION_TIME,
       retry: 0,
     },
   },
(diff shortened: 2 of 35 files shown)


eecavanna:
I'll review this PR later today. I skimmed the code diff on my phone a few days ago and did not notice any issues.

During this morning's squad meeting, I said I wanted the app to at least alert the user before changes they made while they were offline get written to the server (potentially overwriting changes made on the server since then, since the lock isn't being obtained for offline editing). I'll create a separate ticket about that.

src/QueryClientProvider.tsx
```diff
@@ -9,11 +9,13 @@ import { useStore } from "./Store";
 import { createAsyncStoragePersister } from "@tanstack/query-async-storage-persister";
 import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 
+const GARBAGE_COLLECTION_TIME = 1000 * 60 * 60 * 24 * 7; // 1 week
```
eecavanna, line 12: I was going to suggest using a word like "interval" or "duration" instead of "time," given that the value is not a timestamp (but is, instead, a temporal "distance" or "width").

```suggestion
const GARBAGE_COLLECTION_INTERVAL = 1000 * 60 * 60 * 24 * 7; // 1 week
```

Then, I noticed the names of the properties of the `defaultOptions.queries` object—they end with "`...Time`". In that case, although I still prefer "`..._INTERVAL`" here, I see a benefit of keeping this constant's name ending with `..._TIME`".

src/Store.tsx
```diff
@@ -81,22 +83,40 @@ const StoreProvider: React.FC<PropsWithChildren> = ({ children }) => {
...
+        nmdcServerClient.setRefreshToken(refreshToken);
+      }
+
+      // If the user is online, attempt to exchange the refresh token for an access token. If
+      // successful, update the context to reflect the currently logged-in user. If the exchange
+      // fails, clear the logged-in user from the context and storage.
+      // If the user is offline, we don't do anything here. We know the refresh token exchange
+      // will fail so there's no need to attempt it. And we don't want to clear the logged-in user
+      // from the context if it was loaded from storage. This allows a user to be "presumptively"
+      // logged in if the app is opened while offline.
```
eecavanna, line 107: Thank you for including this documentation!

src/Store.tsx
```diff
@@ -123,13 +143,37 @@ const StoreProvider: React.FC<PropsWithChildren> = ({ children }) => {
...
+   * the access token stored in the NmdcServerClient instance. If storage is provided, persist
+   * the user to storage as well.
+   *
+   * @param storage
+   */
+  async function _updateLoggedInUser(storage: Storage | null) {
+    const user = await nmdcServerClient.getCurrentUser();
+    setLoggedInUser(user);
+    if (storage) {
+      await storage.set(StorageKey.LOGGED_IN_USER, user);
```
eecavanna, line 157: I assume `user` is a "JSON object" as opposed to a string. Based on that assumption, I was surprised to see it being passed to `storage.set` without it being "stringified" first. Maybe `storage.set` takes care of that internally.

I see this statement in the [Capacitor docs](https://capacitorjs.com/docs/v3/apis/storage#working-with-json):
> The Storage API only supports string values. You can, however, use JSON if you `JSON.stringify` the object before calling `set()`, then `JSON.parse` the value returned from `get()`.

src/api.ts
```diff
@@ -405,13 +413,16 @@ class NmdcServerClient extends FetchClient {
   // multiple requests that require authentication to be made in quick succession, and if they all
   // fail because of an expired or missing access token we don't want to initiate a token refresh
   // for each request when only one is needed.
-  async exchangeRefreshToken(refreshToken: string) {
+  async exchangeRefreshToken() {
     if (this.exchangeRefreshTokenCache !== null) {
       return this.exchangeRefreshTokenCache;
     }
+    if (this.refreshToken === null) {
+      throw new Error("No refresh token");
```
eecavanna, line 421:
```suggestion
      throw new Error("No refresh token found");
```

src/components/ErrorBanner/ErrorBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+        </IonButton>
+      </Banner>
+      {isErrorDetailShown && (
+        <div className={`ion-padding ${styles.errorDetails}`}>
+          An unexpected error occurred while communicating with the server.
+          Please try again later. If the problem persists, please email{" "}
+          <a
+            target="_blank"
+            rel="noopener noreferrer"
+            href={`mailto:${SUPPORT_EMAIL}?subject=Field Notes Error&body=${emailBody}`}
```
eecavanna, line 52: I wonder whether some kind of URL-escaping will be necessary for some email bodies.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

src/useIsSubmissionEditable.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { isOnline } = useNetworkStatus();
+
+  // If no user is logged in, the submission is not editable
+  if (loggedInUser === null) {
+    return false;
+  }
+
+  // If someone is logged in and is offline, there is no way they can claim the lock so just
+  // assume that no one else is editing. This may result in data clobbering once this user goes
+  // back online.
```
eecavanna, line 21:
```suggestion
  // assume that no one else is editing.
  // 
  // WARNING: This may result in data clobbering once this user goes back online.
```

src/components/MutationErrorBanner/MutationErrorBanner.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { PropsWithChildren } from "react";
+import { UseMutationResult } from "@tanstack/react-query";
+import ErrorBanner from "../ErrorBanner/ErrorBanner";
+
+interface MutationErrorBannerProps extends PropsWithChildren {
+  // Not clear why the `any` can't be replaced by `unknown` here.
```
eecavanna, line 6: Thanks for documenting this (in case a future maintainer feels motivated to try to dig into it)!

src/components/SampleSlotEditModal/SampleSlotEditModal.tsx
```diff
@@ -297,7 +307,14 @@ const SampleSlotEditModal: React.FC<SampleSlotEditModalProps> = ({
           <IonGrid>
             <IonRow>
               <IonCol>
-                <IonButton fill="outline" expand="block" onClick={onCancel}>
+                <IonButton
+                  fill="outline"
+                  expand="block"
+                  // Call dismiss here instead of onCancel directly so that the modal can animate
+                  // out of view before the onCancel handler is called via the IonModalDidDismiss
+                  // event
```
eecavanna, line 315:
```suggestion
                  // Call `dismiss` here instead of `onCancel` directly so that the modal can animate
                  // out of view before the `onCancel` handler is called via the `IonModalDidDismiss`
                  // event.
```

src/components/ThemedToolbar/ThemedToolbar.tsx
```diff
@@ -1,12 +1,37 @@
...
+          <div className={classes.offlineWarning}>
+            <IonIcon icon={cloudOfflineOutline} />
+            <div>You are offline</div>
+          </div>
+        </IonToolbar>
+      )}
+      <IonToolbar {...rest} className={classes.themedBackground}>
+        {children}
+      </IonToolbar>
+      {(isFetching > 0 || isMutating > 0) && (
```
eecavanna, line 30: I was surprised to see these values being compared with a number (i.e. 0), as—based upon their names—I had assumed they were boolean values.

src/mocks/fixtures.ts
```diff
@@ -197,3 +198,27 @@ export const submissions: SubmissionMetadata[] = [
     source_client: "field_notes",
   },
 ];
+
+export class FakeErrorTestClient extends FetchClient {
```
eecavanna, line 202: It's not immediately apparent to me what this class is. I think I'd find a 1-2 line summary of the class helpful (e.g. in a docstring above the class signature).

src/pages/SamplePage/SamplePage.tsx
```diff
@@ -104,6 +103,11 @@ const SamplePage: React.FC = () => {
     updateMutation.mutate(updatedSubmission, {
       onSuccess: () => setModalSlot(null),
     });
+    if (!isOnline) {
+      // If we're offline, the updateMutation will stay in a paused state and not fire onSuccess
+      // until we go back online. So we need to manually close the modal here.
```
eecavanna, line 108:
```suggestion
      // If we're offline, the `updateMutation` will stay in a paused state and not fire `onSuccess`
      // until we go back online. So we need to manually close the modal here.
```

eecavanna (approved):
Great PR description! Thanks for implementing all this. I did not come across anything that I consider to be a deal breaker. I left some suggestions about variable names, comment formatting, and other non-deal-breaker stuff.

src/NetworkStatus.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const NetworkStatusContext = createContext<NetworkStatusContextValue>({
+  isOnline: true,
+});
+
+const NetworkStatusProvider: React.FC<PropsWithChildren> = ({ children }) => {
+  const [isOnline, setIsOnline] = useState<boolean | null>(null);
+
+  useEffect(() => {
+    // Used to prevent setting up the listener from being added from an unmounted component
```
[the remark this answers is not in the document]
pkalita-lbl: It is to avoid having multiple listeners be attached simultaneously. The teardown function returned by the effect will remove the listener _if_ it has been created by the time the teardown function is called. But because the setup function is async, it's possible for the component to be mounted and unmounted all while the async action is still happening.

src/QueryClientProvider.tsx
pkalita-lbl: Point taken, but I think I'll rather stay consistent with TanStack Query's `gcTime` option and all their docs.

src/Store.tsx
pkalita-lbl: Oh yeah, this is all very confusing. 

We're not using a Capacitor plugin for storage. The `@capacitor/storage` plugin (quoted above) was renamed to `@capacitor/preferences` a [few releases ago](https://capacitorjs.com/docs/updating/4-0#storage). And that plugin is designed to be a simple key/value store for "lightweight data". That's why we're using [`@ionic/storage`](https://github.com/ionic-team/ionic-storage) instead.

The `@ionic/storage` docs don't make this incredibly clear, but under the hood they use [`localForage`](https://github.com/localForage/localForage) as a storage backend. And `localForage` does [automatically stringify and parse values](https://github.com/localForage/localForage?tab=readme-ov-file#storing-blobs-typedarrays-and-other-js-objects).

src/components/ErrorBanner/ErrorBanner.tsx
pkalita-lbl: It happens on line 29 😁

src/components/ThemedToolbar/ThemedToolbar.tsx
pkalita-lbl: Me too! Before I added the `> 0` I was confused why the number 0 was appearing in the UI.

src/mocks/fixtures.ts
pkalita-lbl: Done

Issue #167, opened 2024-08-27
Consider whether the web version on GitHub Pages is necessary


pkalita-lbl:
Currently we have a web version of the app hosted by GitHub Pages and accessible at https://fieldnotes.microbiomedata.org/. This was primarily useful to demonstrate work to non-developers before we were able to build and deploy native versions of the app. Now that we can build and distribute native builds, there might not be much of a need for the deployed web version anymore. 

One important caveat is that the static files hosted under the `/.well-known` directory need to continue to be available there. However, that doesn't mean we need to deploy the _full_ app. Nor does it mean we need to even continue to use GitHub Pages to host those files if there is a better option.

So the tasks here are:

- [ ] Work with stakeholders to determine if there is any reasonable use of the web version of the app on https://fieldnotes.microbiomedata.org
- [ ] If yes, consider whether it should be "production-ized" (i.e. running specific tagged versions instead of `main`, pointing to the prod `nmdc-server` API, etc).
- [ ] If no, decommission it while keeping the `/.well-known` files intact.

pkalita-lbl:
See also: https://github.com/microbiomedata/nmdc-field-notes/issues/166

pkalita-lbl:
The web version is occasionally useful in meetings, so we'll keep it around. We did implement the banner warning people to not use it for real.

PR #168, opened 2024-08-28
Add dev site warning banner; display it on GH Pages site

### LINKED ISSUE #166: Add banner to GitHub Pages site indicating that it is not a production site
Currently we have a web version of the app hosted by GitHub Pages and accessible at https://fieldnotes.microbiomedata.org. This was primarily useful to demonstrate work to non-developers before we were able to build and deploy native versions of the app. And because it was a testing/demoing site it was (and still is) connected to the _dev_ `nmdc-server` API. Now we do have native builds of the app that people are testing. They point to the _prod_ `nmdc-server` API. Additionally, since the GitHub Pages versions rebuilds with every commit to `main` it may have changes that haven't been deployed to native versions yet. I don't expect people to be using that site for much of anything, but we should still somehow warn people of those differences if they stumble upon it.

pkalita-lbl:
Fixes #166 

This is just a quick 'n' dirty change to show a warning banner on the GH Pages site. The goal is to warn people about using it while we decide its ultimate fate (#167).

A new environment variable controls whether the warning banner is shown or not. The environment variable does _not_ get set in any of the existing `.env.*` files. Rather it just gets set as the site is built in the GH Pages deploy workflow.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,6 +36,8 @@ jobs:
 
       - name: Build project
         run: npm run build -- --mode staging
+        env:
+          VITE_SHOW_DEV_SITE_WARNING: "true"
 
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v3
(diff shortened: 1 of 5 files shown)


yxu-lanl (approved):

eecavanna (approved):
Looks good to me. Thanks for implementing this. I'll merge it in now. That will make it live on the PWA (website).

PR #171, opened 2024-09-12
Schema fetching and caching

### LINKED ISSUE #155: `SamplePage` shows blank screen if schema data is not available
Currently the `SamplePage` component doesn't render any content when it has no schema data to work off of:

https://github.com/microbiomedata/nmdc-field-notes/blob/a99e5a0b323246614f06b6a324ee13e309e90e6c/src/pages/SamplePage/SamplePage.tsx#L236-L238

I got into this state while working on offline support by:

1. Emptying the TanStack Query cache (this sort of simulates using the app for the first time)
2. Viewing the logged-in home page (now the persistent cache has information about studies)
3. Going offline
4. Navigating to an individual study's sample page

At this point the schema would have been fetched and cached if we were online, but alas.

So I think there are two action items here:

1. Since the schema is a critical piece of server data for the app to have, figure out a way to fetch and cache it sooner and not depend on actually visiting the sample page.
2. In the (hopefully very unlikely) event that somehow a user _still_ ends up on the sample page with no schema data, we should at least show an error message instead of a blank page.

### LINKED ISSUE #156: Improve method of fetching and caching schema data
See also #155. That issue is about _when_ the schema data is fetched. This is more about _how_ it's fetched and cached. But they may be similar enough to work on together.

---

Currently the query we use to fetch the schema has a fairly long `staleTime`. This prevents it from being refetched very often because (a) it's a fairly large request and (b) it doesn't change very often:

https://github.com/microbiomedata/nmdc-field-notes/blob/a99e5a0b323246614f06b6a324ee13e309e90e6c/src/queries.ts#L322-L339

However, this isn't an entirely optimal solution:

1. In between monthly `nmdc-server` releases, there could be lots of requests (up to 1 every 3 days) for a schema that hasn't changed.
2. If a client fetched the schema data _just_ before an `nmdc-server` release, they could be stuck with the old schema and be out of sync with the server for up to 3 days.

I could imagine solving this problem by:

1. Update `nmdc-server` to be able to provide _just_ the schema version number (as opposed to the whole schema). Currently there is an `/api/version` endpoint, but it only provides the `nmdc-server` version number. Perhaps that could be extended to also provide the schema version number.
2. Let the `useSubmissionSchema` have a shorter `staleTime` so that it refetches more frequently. Update the `queryFn` to _first_ check the version endpoint and compare it to the query's cached data. Only if there is no cached data or if the version numbers are different, request the actual schema data. Otherwise, just return the cached data.

pkalita-lbl:
Fixes #155 
Fixes #156 

### Summary

These changes fix two related issues around when and how the submission schema is fetched. The goals are:

* Prefetch the schema _before_ it is needed to prevent certain bad situations while offline. In the event that the schema is still not available when it is needed, show an error message.
* Only fetch the schema when a new version is available.

### Details

Previously the schema information was fetched at the time it was needed via the `useSubmissionSchema` hook call in the `SamplePage` component. The hook is still called there, but now the data it provides is prefetched via a new `DataPrefetcher` component placed near the top-level of the app. See the [TanStack Query docs](https://tanstack.com/query/v5/docs/framework/react/guides/prefetching) for details on their recommended prefetching approaches. 

Assuming that prefetch happened successfully at least once when opening the app online, then the cached prefetched data will be returned by the `useSubmissionSchema` hook while using the app later, even while offline. In the event that _all_ of that somehow goes wrong and the `SamplePage` component ends up with no submission schema information, it will now show an error message instead of a blank page.

Previously the submission schema query had a long `staleTime` (72 hours) in an attempt to not download the two schema files too often (they are somewhat large). This had the downside of not getting an updated schema after a new release for several days. With these changes the submission schema `queryFn` first checks to see if its cached version matches the server's version (via the `/api/version` endpoint). If it does then no further fetching is done, and the cached version is returned. Otherwise (no cached version or mismatched versions) a fetch for the schema files is done.

### Testing

These changes are tricky to test! Following the steps in #155 is a good start. For testing #166, I verified by downgrading my local `nmdc-server` stack to `nmdc-submission-schema` version 10.7.0, using the app, upgrading back to 10.8.0, and verifying that reloading the app caused a redownload of the schema.

### DIFF
--- a/src/pages/SamplePage/SamplePage.tsx
+++ b/src/pages/SamplePage/SamplePage.tsx
@@ -247,6 +247,14 @@ const SamplePage: React.FC = () => {
             </Banner>
           )}
 
+        {!schema.data && !schema.isFetching && (
+          <Banner color="danger">
+            <IonLabel>
+              <b>Error</b> Missing schema information
+            </IonLabel>
+          </Banner>
+        )}
+
         {schema.data && (
           <>
             <SampleView
--- a/src/pages/SamplePage/SamplePage.tsx
+++ b/src/pages/SamplePage/SamplePage.tsx
@@ -250,7 +250,8 @@ const SamplePage: React.FC = () => {
         {!schema.data && !schema.isFetching && (
           <Banner color="danger">
             <IonLabel>
-              <b>Error</b> Schema not found. Please ensure you are connected to the internet and try again.
+              <b>Error</b> Schema not found. Please ensure you are connected to
+              the internet and try again.
             </IonLabel>
           </Banner>
         )}
--- a/src/pages/SamplePage/SamplePage.tsx
+++ b/src/pages/SamplePage/SamplePage.tsx
@@ -250,7 +250,7 @@ const SamplePage: React.FC = () => {
         {!schema.data && !schema.isFetching && (
           <Banner color="danger">
             <IonLabel>
-              <b>Error</b> Missing schema information
+              <b>Error</b> Schema not found. Please ensure you are connected to the internet and try again.
             </IonLabel>
           </Banner>
         )}
--- a/src/queries.ts
+++ b/src/queries.ts
@@ -3,6 +3,7 @@ import {
   hashKey,
   InfiniteData,
   QueryClient,
+  queryOptions,
   useInfiniteQuery,
   useMutation,
   useQuery,
@@ -31,8 +32,11 @@ export const submissionKeys = {
   lock: (id: string) => [...submissionKeys.locks(), id],
   unlocks: () => [...submissionKeys.all(), "unlock"],
   unlock: (id: string) => [...submissionKeys.unlocks(), id],
-  schemas: () => ["schemas"],
-  submissionSchema: () => [...submissionKeys.schemas(), "submission_schema"],
+};
+
+export const schemaKeys = {
+  all: () => ["schemas"],
+  submissionSchema: () => [...schemaKeys.all(), "submission_schema"],
 };
 
 export function addDefaultMutationFns(queryClient: QueryClient) {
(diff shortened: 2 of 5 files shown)


src/pages/SamplePage/SamplePage.tsx
```diff
@@ -247,6 +247,14 @@ const SamplePage: React.FC = () => {
             </Banner>
           )}
 
+        {!schema.data && !schema.isFetching && (
+          <Banner color="danger">
+            <IonLabel>
+              <b>Error</b> Missing schema information
```
eecavanna, line 253: I think some people will read "Missing" as an adjective as opposed to a verb, and then wonder, "What about the missing schema information?" With that in mind, I have this proposal, which I also consider to be actionable:

```suggestion
              <b>Error</b> Schema not found. Please ensure you are connected to the internet and try again.
```

This is not a deal-breaker, from my perspective, though. Also, I don't know how much screen real estate is available to the `<Banner ...>` element. The wording I'm suggesting is several times longer than the original.

src/queries.ts
```diff
@@ -308,12 +314,31 @@ export function useSubmissionCreate() {
...
+      );
+      if (cachedData !== undefined) {
+        // We have a copy of the schema in the cache. Check to see if it is still the latest version
+        const versionInfo = await nmdcServerClient.getVersionInfo();
+        // If the cached schema is the same version as the current server version, return the cached
+        // schema
+        if (cachedData.schema.version === versionInfo.nmdc_submission_schema) {
+          return cachedData;
+        }
+      }
```
eecavanna, line 340: Non-deal-breaker nit-pick: I moved the comments to above the `if` block to contrast it with the `// We don't have a ...` block that follows the `if` block.

```suggestion
      // If we have a cached schema and it's the same version as the server
      // would give us if we were to fetch it now, return that cached schema.
      if (cachedData !== undefined) {
        const versionInfo = await nmdcServerClient.getVersionInfo();
        if (cachedData.schema.version === versionInfo.nmdc_submission_schema) {
          return cachedData;
        }
      }
```

eecavanna (approved):
Looks good to me! Helpful PR description! I would be comfortable with this branch being merged in as-is.

I left a couple suggestions about things that I do not consider to be deal-breakers.

src/pages/SamplePage/SamplePage.tsx
pkalita-lbl: > I don't know how much screen real estate is available to the <Banner ...> element. 

It will grow to fit whatever message you give it. In this case there wouldn't be anything else on the page so there's no harm in being more verbose here.

Issue #182, opened 2024-09-18
Wrong build number displayed in app for build `0.1.0 (5)`


pkalita-lbl:
### Bug summary

Wrong build number displayed in app for build `0.1.0 (5)`

### Bug details

After installing build `0.1.0 (5)`, the Settings page still displays `0.1.0 (4)` in the Version field. It should say `0.1.0 (5)`.

### App version

0.1.0 (5)

### Device type

None

### Operating system version

Confirmed on both iOS and Android

pkalita-lbl:
Currently we have in our `package.json` `scripts`:

```json
    "prerelease": "vitest run && npm run build.native",
    "release": "./scripts/release.sh",
```

I can't remember the rationale for having `npm run build.native` as part of the `prerelease` step. It seems like it would make more sense to be in `postrelease`. That's just my gut reaction, but it needs a closer look.

ssarrafan:
@pkalita-lbl who should this be assigned to?

pkalita-lbl:
It's kind of up-for-grabs for any of the Field Notes developers. But if you're looking to have all the current sprint issues assigned to someone I'll take it.

PR #192, opened 2024-10-02
Run ionic build after trapeze in release script

pkalita-lbl:
Fixes https://github.com/microbiomedata/nmdc-field-notes/issues/182

As pointed out in the issue, there is a `prerelease` script which runs `npm run build.native` before running the `release.sh` script:

https://github.com/microbiomedata/nmdc-field-notes/blob/d862c821a1a696b0c217b1a55b39d3599108016c/package.json#L24-L25

The purpose of that build is to ensure that we didn't accidentally forget to sync the native projects and commit those changes before making a release (`release.sh` will stop if there are uncommitted changes leftover from that build). We still want to do that, so I've kept that in place.

However, we do need _another_ build _after_ the new build/version numbers are set. This allows the new numbers to be baked into the web assets that the native projects use. I've added that to the `release.sh` script. 

This means that when you run `npm run release` _two_ builds will happen (before _and_ after incrementing the build/version numbers). Maybe not ideal, but I'm not sure I see a way around it. And the builds don't take that long anyway.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -49,11 +49,15 @@ else
   log "New version number: $VERSION_NUMBER"
 fi
 
-# Run trapeze to propagate the new version and build number to native projects
+# Run trapeze to propagate the new version and build number to native project files
 log "Updating native projects"
 export FIELD_NOTES_VERSION_NUMBER=$VERSION_NUMBER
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
The changes and overall procedure make sense to me. Thanks for taking care of this (resolving the issue). Also, thanks for explaining the situation, via the PR description.

PR #193, opened 2024-10-02
Check `any_of` on slot definition when determining input type

### LINKED ISSUE #176: ENVO dropdowns are not available
### Bug summary

Envo terms are not available as dropdowns in app

### Bug details

I expected the ENVO categories/terms to be present as some sort of selectable list such as the options for growth facility or analysis/data type slots when I click the relevant slots (e.g., "broad scale environmental context"). I don't know the ENVO terms off the top of my head so I was not able to format them while out in the field and would have to do this later on the Submission Portal. As required fields, this prevents me from fully validating my metadata in-app.

### Steps to reproduce

In a sample, select broad-scale environmental context, local environmental context, or environmental medium. Pop-up is text-entry only, not a list of options.

### Anything else?

_No response_

### App version

0.1.0

### Device type

Apple

### Operating system version

iOS 17.6.1

pkalita-lbl:
Fixes #176 

These changes update the helper method (`getSelectState`) in `src/components/SampleSlotEditModal/SampleSlotEditModal.tsx` which determines whether to show a select box or a plain text box for a given slot. Previously the logic only looked at the slot's `range`. That's sufficient most of the time, but there are a few slots that (mainly for historical reasons) allow either a string _or_ an enum (via the `any_of` slot attribute). Previously `getSelectState` would return `isSelectable: false` in that case. Now it will return `isSelectable: true` and the appropriate permissible values.

### DIFF
--- a/src/components/SampleSlotEditModal/SampleSlotEditModal.test.ts
+++ b/src/components/SampleSlotEditModal/SampleSlotEditModal.test.ts
@@ -1,0 +1,90 @@
+import { getSelectState } from "./SampleSlotEditModal";
+import { SchemaDefinition } from "../../linkml-metamodel";
+import { GoldEcosystemTreeNode, SampleDataValue } from "../../api";
+
+const MOCK_SCHEMA: SchemaDefinition = {
+  id: "http://example.org/test",
+  name: "test",
+  slots: {
+    a_string: {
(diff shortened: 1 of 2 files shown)


eecavanna (approved):
LGTM!

PR #205, opened 2024-12-02
Replace individual autofill buttons on name and ORCID with single button

### LINKED ISSUE #198: Replace the magic wand icons in PI name and ORCID with more explicit action button
Some beta feedback indicated that there was confusion about what the magic wand icons in the Name and ORCID fields of the Study create/edit form were for

Image

A lingering concern I've had since we implemented those icons was that it implicitly encourages users to enter their own name and ORCID iD in those fields _even if they aren't the PI of the study_.

One possible way to resolve both issues would be to:

1. Remove the magic wand icons
2. Add additional help text under the "Principal Investigator" subheading that emphasizes we're looking for information about the study PI in this section
3. Add a button with a more explicit label (something like "I am the study's PI"). When that button is clicked fill in the Name and ORCID iD fields with the logged-in user details

pkalita-lbl:
Fixes #198 

### Summary

Previously the Name and ORCID iD fields of the Study form individually had autofill buttons. These changes replace those two buttons with a single autofill button.

### Details

This also includes minor changes to make padding and margins more uniform in the Study form.

Additionally, with the changes in https://github.com/microbiomedata/nmdc-server/pull/1442, the backend now associates an optional email address with User records. These changes also add the `email` field to the app's User model. If the `email` field of the logged in user is present and populated, it will be used to fill out the Study form's email field when the autofill button is pressed. 

| Before Clicking | After Clicking |
| --- | --- |
| image  | image |

### DIFF
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -1,8 +1,9 @@
 import React from "react";
 import {
   IonButton,
-  IonIcon,
   IonInput,
+  IonItem,
+  IonLabel,
   IonSelect,
   IonSelectOption,
   IonText,
@@ -13,7 +14,6 @@ import SectionHeader from "../SectionHeader/SectionHeader";
 import { SubmissionMetadataCreate, TEMPLATES } from "../../api";
 import { Controller, useForm } from "react-hook-form";
 import { useStore } from "../../Store";
-import { colorWand as autoFill } from "ionicons/icons";
 import { StepType } from "@reactour/tour";
 import styles from "./StudyForm.module.css";
 import { useAppTour } from "../AppTourProvider/hooks";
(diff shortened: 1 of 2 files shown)


src/components/StudyForm/StudyForm.tsx
```diff
@@ -87,6 +87,17 @@ const StudyForm: React.FC<StudyFormProps> = ({
       disabled,
     });
 
+  const handlePIAutoFill = () => {
+    if (loggedInUser === null) {
+      return;
+    }
+    setValue("metadata_submission.studyForm.piName", loggedInUser.name);
+    if (loggedInUser.email) {
```
eecavanna, line 95: Note to self/reviewers (no action needed): This is where the app checks whether the email address of the logged-in user is `null` or `""` (or any other [Falsy value](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)). The auto-fill will only happen if the email address is not such a value.

eecavanna (approved):
This looks good to me. I'm glad to see their email address (if present) will be auto-filled also. I'm comfortable with this PR branch being merged in as is.

Issue #177, opened 2024-09-16
Account for changes to `packageName` field on backend


pkalita-lbl:
Part of the proposed changes for https://github.com/microbiomedata/nmdc-server/issues/1367 is changing the `packageName` field of a submission object from a string to an array of strings. This will break certain things in the app. We need to sort out:

1. How the app will handle this change. What would it look like to fully support creating, displaying, and updating studies that use multiple environment templates? Is it simpler to continue to only support a single environment template in the app?
2. How will we coordinate releasing app changes along with the backend changes? Is it possible to update the app to work with _either_ a string or array of strings in `packagename` and release that change _ahead_ of the backend changes?

pkalita-lbl:
Cross-linking the proposed backend changes here: https://github.com/microbiomedata/nmdc-server/pull/1372

ssarrafan:
@pkalita-lbl who should this one be assigned to?

pkalita-lbl:
The work on this is paused until https://github.com/microbiomedata/nmdc-server/issues/1367 is resolved. I believe that work is paused until after the Berkeley refactor roll-out is complete. Moving to next sprint.

ssarrafan:
Berkeley was rolled out so can this be closed @pkalita-lbl ?

pkalita-lbl:
This cannot be closed.

It doesn't really have anything to do with Berkeley. The gating factor is the PR for https://github.com/microbiomedata/nmdc-server/issues/1367 being merged in.

pkalita-lbl:
Still paused waiting for https://github.com/microbiomedata/nmdc-server/issues/1367. Moving to next sprint.

pkalita-lbl:
Still paused waiting for https://github.com/microbiomedata/nmdc-server/issues/1367. Moving to next sprint.

ssarrafan:
@pkalita-lbl update from @marySalvi  on issue #1367: 
_Yes, this will be ready for the December release. This has been blocked by the bug in the runtime code (my runtime code). I have made good progress towards a resolution and am hoping to wrap this up shortly._

Based on this can I move this to the next sprint?

pkalita-lbl:
I moved it.

Issue #203, opened 2024-11-25
Release new beta build before current one expires


eecavanna:
The current build expires in about 22 days (from November 25; so, in mid-December).

ssarrafan:
@eecavanna did this get updated this week?

eecavanna:
No, we did not create a new build of the app this week. I'll move this to the next sprint. We have until mid-December to do this.

pkalita-lbl:
Build `0.1.0 (6)` was released today.

PR #206, opened 2024-12-03
Store slot visibility settings on a per-Study basis

### LINKED ISSUE #150: Sharing custom field visibility options within a study
This is a feedback / user request coming from the GSC workshop, probably a low-priority feature request but I wanted to create an issue while still fresh in my mind :-). It seems like there is (a large) interest for users to toggle field visibility, then share these choices with others in the same study. The use case is a PI with a group of undergrads who will do the actual sampling: their "ideal" process would be (i) they create a study, and pre-select fields to be visible in the App, (ii) the students go into the field, and only see these fields to fill in so are not overwhelmed / have less chances of making mistakes.

Thinking (briefly) about this, I feel one way to do this would be to allow for these field visibility options to be stored and shared at the study level ? So my thoughts/questions for now would be:
- 1: Could we store custom field visibility options at the study level, and set them available for all users when in this study ?
- 2: Could we store multiple custom field visibility options per extension/package ? (may be complex, but I could imagine users asking for this as soon as 1 is implemented, could be fine to say "no we can't")
- 3: Could we export / import multiple custom field visibility options across studies ? (same as 2, I could imagine it being useful, but also not a major barrier, worst case scenario PIs just have to redo the initial selection for their new study, should be reasonable).

pkalita-lbl:
Fixes #150 

### Summary

Currently, slot visibility is stored as an app-level setting. This means that, for a given user, _all_ studies that use the soil template (for example) will show the same set of slots. But two users with access to the same study may see different slots depending on their app settings. During beta testing we received feedback that:

* Users found it confusing that they weren't asked to customize the slot visibility on a per-study basis
* Users wondered how they might share the slot visibility settings with others who have access to the same study

To address those pain points, these changes update slot visibility so that it is a property of a study (technically a `SubmissionMetadata` record), as opposed to a local app setting. The supporting backend changes have already been merged into `nmdc-server` in https://github.com/microbiomedata/nmdc-server/pull/1473.

These changes also address the feedback that choosing from a long list of slots associated with a given template can be overwhelming. The solution here is to use the results of some manual curation of slots by team members to group and prioritize slots that are often collected at sampling time (and thus while using the app) versus slots that might be collected at sampling time vs all other slots.

### Details

* In `src/Store.tsx`, functions for setting and getting slot visibility state have been removed since that information will no longer be stored there.
* The dedicated settings page for slot visibility (`src/pages/FieldVisibilitySettingsPage/FieldVisibilitySettingsPage.tsx`) has been removed, along with the routing and links to the that page (`src/Router.tsx`, `src/components/SettingsAppearanceList/SettingsAppearanceList.tsx`, `src/paths.ts`)
* The `StudyView` component (`src/components/StudyView/StudyView.tsx`) now explicitly lists which template was selected for the study, along with the number of visible slots chosen. The `StudyView` is also now responsible for presenting the `SlotSelectorModal` component when tapping on the template name. This design decision has an eye toward the fact that more than one template will be allowed [in the near future](https://github.com/microbiomedata/nmdc-field-notes/issues/177). Previously, the `SlotSelectorModal` was presented by the `SampleView` component. Since the visible slots are now managed at the study level, this list of slots is passed to `SlotSelectorModal` by `StudyView`.
* After creating a new study, we want the user to immediately be presented the slot visibility selector modal for the newly created study. The idea is to make this feel to the user like a "second step" in the study creation process (even though technically the study is already created at that point). This is accomplished by using a bit of routing state, managed by two new hooks. `useNavigateWithState` exposes a `navigate` function which uses a [`history` object](https://v5.reactrouter.com/web/api/history) to navigate to a new route with state. Then `useNavigationState` is used to access and optionally clear that state. The `StudyCreatePage` component uses `useNavigateWithState` to navigate to the newly created study page with a `openSlotSelectorModalOnEnter` flag set. Then the `StudyViewPage` component uses `useNavigationState` to get that flag and pass it along to the `StudyView` component. The flag is also cleared after it is used so that navigating away from the `StudyViewPage` and then back to it does not re-trigger the slot selector modal.
* Slot grouping and sorting functions (`groupTemplateSlots` in `src/components/SlotSelectorModal/SlotSelectorModal.tsx` and `sortSlots` in `src/utils.ts`) and their usage have been updated to reflect that we are **not** using the grouping and sorting indicated by the submission schema itself anymore, which was designed for the Submission Portal. Instead, we are using a more simplified grouping and sorting that is more user-friendly for the Field Notes app. The file `src/components/SlotSelectorModal/slotVisibilities.ts` was created as a result of a manual assessment of slots by NMDC team members and is used to drive the new grouping logic. 
* Minor tangential changes:
  * A new set of custom CSS variables (`--nmdc-font-size-*`) have been added in order to standardize font sizes. These are also wrapped in new utility CSS classes (`.nmdc-text-*` in `src/theme/global.css`)
  * TypeScript types of variables that are supposed to be template names (i.e. a key of the `TEMPLATES` object in `src/api.ts` -- as opposed to an arbitrary string) have been tighten up.


pkalita-lbl:
This is open for review now even though it is still a draft.

I'm leaving it as a draft as a reminder to myself to not merge it yet. I want to first merge https://github.com/microbiomedata/nmdc-field-notes/pull/205, then put out a new build to TestFlight (https://github.com/microbiomedata/nmdc-field-notes/issues/203), and then merge this.

pkalita-lbl:
Oops, I see I have conflicts to fix. I will get to that later today.

eecavanna:
I'm planning on reviewing this today.

src/components/SlotSelectorModal/slotVisibilities.ts
```diff
@@ -0,0 +1,453 @@
...
+interface SlotVisibilities
+  extends Partial<Record<TemplateName, VisibilityLevel>> {
+  _default?: VisibilityLevel;
+}
+
+/**
+ * This object encodes which slots are commonly, occasionally, or uncommonly measured at sample
+ * collection time. This was the output of a manual curation process. At some point we might
+ * consider encoding this directly in the submission schema somehow. But for now, this is a simple
+ * way to get the job done. The structure of the object is:
```
eecavanna, line 14:
```ts
/**
 * [...] But for now, this is a simple
 * way to get the job done. [...]
 */
```

I'm OK with this file being merged in as is, with that framing of the situation.

One thing that came to mind while looking at the `slotVisibilities` object below was that those three strings could be implemented as [Enum](https://www.typescriptlang.org/docs/handbook/enums.html) values rather than literal strings, which would allow for the inner objects to look like this:

```diff
  agrochem_addition: {
-    _default: "occasional",
+    _default: Frequency.Occasional,
  },
```

That might complicate whatever automatic process was used to convert the Google Sheets document into this object, though.

src/components/StudyView/StudyView.tsx
```diff
@@ -150,6 +213,34 @@ const StudyView: React.FC<StudyViewProps> = ({ submissionId }) => {
...
+                    {item.visibleSlots === undefined ? (
+                      "Not customized"
+                    ) : (
+                      <>
+                        <Pluralize
+                          count={item.visibleSlots.length}
+                          singular={"field"}
+                          showCount
+                        />{" "}
+                        chosen
```
eecavanna, line 235: I don't know exactly what this message looks like in context. Another word that came to mind is "visible" (as an alternative to "chosen").

src/components/StudyView/StudyView.tsx
```diff
@@ -159,6 +250,16 @@ const StudyView: React.FC<StudyViewProps> = ({ submissionId }) => {
                 : undefined
             }
           />
+
+          {/* TODO: lock/unlock submission when opening/closing the modal */}
```
eecavanna, line 254: I will assume this is a `TODO` for the future (after this branch gets merged in) as opposed to before this branch gets merged in.

eecavanna (approved):
Looks good! I found the PR description very helpful, both to understand the real-world context (i.e. motivation for doing this) and to understand various aspects of the diff.

I left some comments, including at least one that is to remind myself to look into something. I will re-read and act upon that/those right after I submit this review. _Edit: I'm done doing that._

src/components/SlotSelector/SlotSelector.tsx
```diff
@@ -105,6 +104,12 @@ const SlotSelector: React.FC<SlotSelectorProps> = ({
         </IonCheckbox>
       </SectionHeader>
       <IonList className="ion-padding-bottom">
+        {group.description && (
+          <div
+            className="ion-padding-horizontal ion-padding-bottom nmdc-text-sm"
+            dangerouslySetInnerHTML={{ __html: group.description }}
```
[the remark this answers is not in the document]
eecavanna: Update: I have an idea for an alternative. Instead of storing `group.description` as an arbitrary string, store it as a React node. Then, instead of rendering the string in the current way (i.e. via `dangerouslySetInnerHTML`), render the equivalent React node.

Here's what the definition of `uncommonGroup`, for example, would look like:

```tsx
  const uncommonGroup: SlotGroup = {
    name: "uncommon",
    description: <>
      These fields are rarely measured at the time of sample collection.
      These field are more often entered after sampling via the
      <a target="_blank" rel="noopener noreferrer" href="https://nmdc-documentation.readthedocs.io/en/latest/tutorials/submission_portal.html">
        NMDC Submission Portal
      </a>.
    </>,
    title: "Uncommon",
    slots: [],
  };
```

> The definition of `SlotGroup` may also require a change in order to accommodate this.

Here's what the rendering code would look like:

```tsx
<div className="ion-padding-horizontal ion-padding-bottom nmdc-text-sm">
    {group.description}
</div>
```

src/api.ts
```diff
@@ -180,10 +181,15 @@ export interface SubmissionMetadataCreate extends SubmissionMetadataBase {
 export interface SubmissionMetadataUpdate extends SubmissionMetadataBase {
   id: string;
   status?: string;
+  field_notes_metadata?: Nullable<FieldNotesMetadata>;
   // Map of ORCID iD to permission level
   permissions?: Record<string, string>;
 }
 
+export interface FieldNotesMetadata {
+  fieldVisibility?: Partial<Record<TemplateName, string[]>>;
```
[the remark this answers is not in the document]
pkalita-lbl: Will do.

I like the idea of introducing a `SlotName` type. It wouldn't necessarily add anything strictly from a type-checking perspective (I don't see us going to the trouble of maintaining something like `type SlotName = "collection_time" | "depth" | ...`), but it does provide a nice semantic hint.

src/components/SlotSelector/SlotSelector.tsx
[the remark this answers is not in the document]
pkalita-lbl: Oh, yeah, duh that's a much better idea!

src/components/SlotSelectorModal/slotVisibilities.ts
pkalita-lbl: > whatever automatic process was used to convert the Google Sheets document into this object

Haha, if only. No, it was a manual process. So no worries (or lots of worries??) there. 

The values in this object are already constrained via a string union type: https://github.com/microbiomedata/nmdc-field-notes/blob/b56760aba47238449d06859107484939a57e2c67/src/components/SlotSelectorModal/slotVisibilities.ts#L3

I couldn't tell you in detail the pros/cons of string unions vs enums in a situation like this. I guess I lean toward the string unions as being a bit more lightweight.

src/components/StudyView/StudyView.tsx
pkalita-lbl: I went with chosen here because even though the words visible and visibility are all over the code, they don't really appear much in the final interface.

The text in the modal itself actually starts with the imperative "Select..." so maybe "selected" is even better here.

src/components/StudyView/StudyView.tsx
pkalita-lbl: Correct

PR #208, opened 2024-12-10
Add support for multiple environmental templates

### LINKED ISSUE #177: Account for changes to `packageName` field on backend

pkalita-lbl:
Fixes #177 

### Summary

These changes add support for selecting multiple environmental templates. This is enabled by backend support provided by https://github.com/microbiomedata/nmdc-server/pull/1372 and https://github.com/microbiomedata/nmdc-server/pull/1491.

Because we can't 100% control the timing of the app's release relative to the backend's release, the overall approach of these changes is to make the app agnostic to the `packageName` field being a single string (the old single template format) _or_ an array of strings (the new multiple template format).

The one thing that the app can't know _a priori_ is whether it should send a single string or an array of strings when creating a new submission (if you're working with an existing submission the app will just keep whatever it was given -- string or array -- by the backend). The one place that's controlled is in `src/data.ts` and that's what we'll need to change once the backend changes are on production and we're ready to have multiple environment selection be the default.

### Details

* Most of the logic to normalize `packageName` as a string vs an array of strings is encapsulated in new function in `src/utils.ts`. 
  * `getSubmissionTemplates` is new and returns an array of template names regardless of whether the submission's `packageName` field is a string or an array of strings.
  * `getSubmissionSamples` previously returned a list of sample metadata objects. Now it returns an object where each key is a template name and the value is a list of sample metadata objects for that template.
  * `getSubmissionSamplesCount` adds up the number of sample metadata objects across all of the templates.
  * `getSubmissionSamplesForTemplate` is new and is mainly a shortcut for indexing into the object returned by `getSubmissionSamples`. 
  * `getSubmissionSample` previously only needed an numeric index to identify a single sample within a submission. Now it needs both a template name _and_ a numeric index.
* Because samples are now identified by their template plus a numeric index, the `SamplePage` now requires a `template` parameter. This is also reflected in the routing to this page (see: `src/Router.tsx`, `src/paths.ts`, `src/components/SampleList/SampleList.tsx`)
* The `SampleList` component still displays all samples in a single flat list. This means it needs to do a little extra work to keep track of a "flattened index" for the purpose of uniquely identifying samples _in that list_. It also needs to ask the user which template to use when creating a new sample (if more than one template is associated with the study). And finally the `IonChip` elements it presents are now interactive, as a way to filter the samples by template.
* When editing a study (`src/components/StudyForm/StudyForm.tsx`) there was existing logic to keep two _backend_ data fields (`packageName` and `templates`) in sync (arguably the backend should do this automatically, but it doesn't for reasons perhaps lost to history). This logic got a bit more complex in order to handle both plain strings and arrays of strings.
* `StudyView` (`src/components/StudyView/StudyView.tsx`) has been updated so that it automatically shows the slot selector modal for _each_ template where the user has yet to choose visible slots for that template. This actually removes a bit of the logic based on routing state introduced in #206. After thinking about it more, I realized we should probably just prompt the user to choose their visible slots if they haven't done so already no matter how they landed on that page. Mainly I'm thinking about the use case where someone creates a submission in the Submission Portal and then starts working with that submission in the app. We wouldn't want them to miss the slot visibility selection step just because they didn't create the submission in the app.
* There are a number of places where I'm just standardizing the app's terminology and variable naming around the term "template". Previously, there were some places where the same concept was called "package name" in the code. That is what the backend calls it (again, reasons lost to time), but I wanted the app's code to at least be consistent with itself.

### Testing

The most important thing is that these change should work while talking to a backend running the latest `nmdc-server` changes (i.e. running the `main` branch locally or pointing to `data-dev.microbiomedata.org`) _or_ the current production `nmdc-server` version (i.e. running the `v1.1.1` tag locall or pointing to `data.microbiomedata.org`).

Using either backend you should be able to create new submissions, edit existing ones, and add samples to a submission.

You will see slightly different behaviors when using the dev vs. prod backends. When using dev you should see the templates explicitly listed on the study view page, and tapping on one should bring up the slot selection modal. You will not see on that prod because that version of the backend does not support it yet.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -82,7 +82,9 @@ export interface SampleData {
 }
 
 export interface IndexedSampleData extends SampleData {
-  _index: number;
+  _flatIndex: number;
+  _templateIndex: number;
+  _template: TemplateName;
 }
 
 // This should eventually come from the schema itself
@@ -154,7 +156,7 @@ export type TemplateName = keyof typeof TEMPLATES;
 export type SlotName = string;
 
 export interface MetadataSubmission {
-  packageName: TemplateName | "";
+  packageName: TemplateName | TemplateName[] | "";
   contextForm: ContextForm;
   addressForm: AddressForm;
   templates: string[];
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -11,7 +11,7 @@ import {
 } from "@ionic/react";
 import RequiredMark from "../RequiredMark/RequiredMark";
 import SectionHeader from "../SectionHeader/SectionHeader";
-import { SubmissionMetadataCreate, TEMPLATES } from "../../api";
+import { SubmissionMetadataCreate, TemplateName, TEMPLATES } from "../../api";
 import { Controller, useForm } from "react-hook-form";
 import { useStore } from "../../Store";
 import { StepType } from "@reactour/tour";
@@ -283,7 +283,10 @@ const StudyForm: React.FC<StudyFormProps> = ({
                       "metadata_submission.templates",
                       newPackageName.concat(
                         templates.filter(
-                          (template) => !previousPackageName.includes(template),
+                          (template) =>
+                            !previousPackageName.includes(
+                              template as TemplateName,
+                            ),
                         ),
                       ),
                     );
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -79,7 +79,7 @@ const StudyForm: React.FC<StudyFormProps> = ({
 
   const { loggedInUser } = useStore();
 
-  const { handleSubmit, control, formState, setValue } =
+  const { handleSubmit, control, formState, setValue, getValues } =
     useForm<SubmissionMetadataCreate>({
       defaultValues: submission,
       mode: "onTouched",
@@ -257,11 +257,38 @@ const StudyForm: React.FC<StudyFormProps> = ({
                 data-tour={`${TourId.StudyForm}-2`}
                 className={`${(fieldState.isTouched || formState.isSubmitted) && "ion-touched"} ${fieldState.invalid && "ion-invalid"}`}
                 labelPlacement="floating"
+                multiple={Array.isArray(field.value)}
                 onIonDismiss={field.onBlur}
                 onIonChange={(e) => {
-                  // these two fields need to stay in sync
-                  setValue("metadata_submission.templates.0", e.detail.value);
-                  field.onChange(e.detail.value);
+                  // The `packageName` and `templates` fields need to stay in sync.
+                  const previousPackageName = getValues(
+                    "metadata_submission.packageName",
+                  );
+                  const newPackageName = e.detail.value;
+                  // The value of the `packageName` field is typed to be a string or an array of
+                  // strings to ease the transition between the formats that the backend accepts.
+                  // Once the transition is complete, this logic can be simplified to only use
+                  // the array format.
+                  if (typeof previousPackageName === "string") {
+                    // If the previous value was a string, just replace the first element of
+                    // the templates array with the new package name.
+                    setValue("metadata_submission.templates.0", newPackageName);
+                  } else if (Array.isArray(previousPackageName)) {
+                    // If the previous value was a string, remove all the previous package names
+                    // from the templates array and prepend the new package names.
+                    const templates = getValues(
+                      "metadata_submission.templates",
+                    );
+                    setValue(
+                      "metadata_submission.templates",
+                      newPackageName.concat(
+                        templates.filter(
+                          (template) => !previousPackageName.includes(template),
+                        ),
+                      ),
+                    );
+                  }
+                  field.onChange(newPackageName);
                 }}
                 {...field}
               >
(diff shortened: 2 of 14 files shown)


src/components/StudyForm/StudyForm.tsx
```diff
@@ -257,11 +257,41 @@ const StudyForm: React.FC<StudyFormProps> = ({
                 data-tour={`${TourId.StudyForm}-2`}
                 className={`${(fieldState.isTouched || formState.isSubmitted) && "ion-touched"} ${fieldState.invalid && "ion-invalid"}`}
                 labelPlacement="floating"
+                multiple={Array.isArray(field.value)}
                 onIonDismiss={field.onBlur}
                 onIonChange={(e) => {
-                  // these two fields need to stay in sync
-                  setValue("metadata_submission.templates.0", e.detail.value);
-                  field.onChange(e.detail.value);
+                  // The `packageName` and `templates` fields need to stay in sync.
```
eecavanna, line 263: As I am reading this (before having read the remainder of the diff), I am wondering whether the author meant to write `template`—instead of `templates`—in the comment.

_Edit: Nope! It's a reference to the `templates` (plural) property of the `metadata_submission` object. I'll resolve this conversation immediately._

src/components/StudyView/StudyView.tsx
```diff
@@ -51,29 +50,45 @@ const StudyView: React.FC<StudyViewProps> = ({
...
     }
     return fieldVisibilityInfo;
   }, [submission.data]);
 
-  useIonViewDidEnter(() => {
-    if (openSlotSelectorModalOnEnter) {
-      setModalTemplateVisibleSlots(templateVisibleSlots[0]);
+  useEffect(() => {
+    // If the slot selector is not already open, iterate through the study's templates. If one of
+    // them has no visible slot information (this could be because it is a brand-new study or
```
eecavanna, line 74: I felt confused when I first read this comment (reading it as "If the slot information of one of the templates is not visible..."). I find this alternative to be more clear:

```suggestion
    // them has no slot visibility information (this could be because it is a brand-new study or
```

src/data.ts
```diff
@@ -68,7 +68,7 @@ export const initMetadataSubmission = (): MetadataSubmission => ({
   addressForm: initAddressForm(),
   contextForm: initContextForm(),
   multiOmicsForm: initMultiOmicsForm(),
-  packageName: "",
+  packageName: "", // TODO: CHANGE DEFAULT WHEN BACKEND IS UPDATED
```
eecavanna, line 71: I think this alternative would be more instructive to other maintainers, but I am not sure it accurately represents the original author's intent.

```suggestion
  packageName: "", // TODO: CHANGE DEFAULT VALUE TO EMPTY ARRAY WHEN BACKEND IS UPDATED
```

src/utils.ts
```diff
@@ -1,41 +1,141 @@
...
   }
-  const environmentalPackageName = submission.metadata_submission.packageName;
-  const sampleDataField = environmentalPackageName
-    ? TEMPLATES[environmentalPackageName].sampleDataSlot
-    : undefined;
-  if (!sampleDataField) {
+  let count = 0;
+  const templates = getSubmissionTemplates(submission);
+  templates.forEach((template) => {
+    const sampleDataSlot = TEMPLATES[template]?.sampleDataSlot;
```
eecavanna, line 85: This is a follow-up comment to the first one I left. This is the final file in the diff. After having reviewed the other files in the diff, I'd prefer `template` be named `templateName`. Given what `template` contains (i.e. the name of a template), I find `TEMPLATES[templateName]` easier to mentally process than `TEMPLATES[template]`.

src/api.ts
```diff
@@ -82,7 +82,9 @@ export interface SampleData {
 }
 
 export interface IndexedSampleData extends SampleData {
-  _index: number;
+  _flatIndex: number;
+  _templateIndex: number;
+  _template: TemplateName;
```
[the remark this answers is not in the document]
eecavanna: After having read the rest of the diff, I would prefer the various `template` variables that contain template names, be named `templateName`; to distinguish them from the object stored in `TEMPLATES[templateName]`, which I assume is a "template".

eecavanna (approved):
Thanks for implementing this, including coordinating the backend changes. I found the PR description very helpful in preparing my mind with relevant context for reviewing the diff.

I'm comfortable with this being merged in as is. I have one concern that I do not consider to be a deal breaker. I think resolving it would involve making a large quantity of small changes to the diff. That is: to refer to a template name as `templateName` instead of as `template`. I left some comments on the diff about this.

I have not tested this branch on my system yet. I will do that now. I don't think I'll have time to exercise all the new functionality, but I will do some baseline tests.

eecavanna:
In my local development environment, I failed to test this against the production `nmdc-server`. I ran into an issue during ORCID Login (unrelated to this branch).

image

With that being the case, I will switch to testing against the development `nmdc-server`.

src/api.ts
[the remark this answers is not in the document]
pkalita-lbl: Yeah I like the idea. I started doing that and then realized that it would actually be _a lot_ of changes; enough that I'd be worried about breaking things. So I think I might save that as a follow-on task.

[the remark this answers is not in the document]
pkalita-lbl: https://github.com/microbiomedata/nmdc-field-notes/issues/209

PR #211, opened 2024-12-18
Enable multiple templates by default

pkalita-lbl:
Now that Release 2024.12 is on production, these changes complete the corresponding app changes by enabling multiple environment selection for new studies.

I also made a small unrelated tweak to the theme stylesheet so that list items don't have an award gray background on Android in dark mode. These changes ensure that list items always have the same background color as the main interface no matter what platform or light/dark mode.

### DIFF
--- a/src/components/StudyForm/StudyForm.tsx
+++ b/src/components/StudyForm/StudyForm.tsx
@@ -34,8 +34,8 @@ const steps: Array<StepType> = [
     selector: `[data-tour="${TourId.StudyForm}-2"]`,
     content: (
       <>
-        Here, you can select a template that reflects the kind of environment
-        you&apos;ll be collecting samples from.
+        Here, you can select the templates that reflects the kind of
+        environments you&apos;ll be collecting samples from.
       </>
     ),
   },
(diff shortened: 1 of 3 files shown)


src/components/StudyForm/StudyForm.tsx
```diff
@@ -34,8 +34,8 @@ const steps: Array<StepType> = [
     selector: `[data-tour="${TourId.StudyForm}-2"]`,
     content: (
       <>
-        Here, you can select a template that reflects the kind of environment
-        you&apos;ll be collecting samples from.
+        Here, you can select the templates that reflects the kind of
```
eecavanna, line 37: I think there was a disagreement between the plural "templates" and the "reflects."

I also changed "kind" to "kinds" because I thought each checkbox represented one "kind" of environment, and they could now select multiple "kinds."

```suggestion
        Here, you can select the templates that reflect the kinds of
```

pkalita-lbl: Oh yeah, good catch

PR #215, opened 2025-01-06
Issue 213 gps coord digits

### LINKED ISSUE #213: Autofill GPS coordinates not validating
### Bug summary

With latest version, the "Use GPS location" fills in current lat/lon, but indicates "Value does not match pattern."

### Bug details

Expected to have the autofill function populate the current GPS coordinates, but comes up with error. I first selected a "one time" GSP location. Here is a screenshot of what I see. 
IMG_8370170DE7DA-1

### Steps to reproduce

Start a new sample, use plant-associated template, click "Use GPS location".

### Anything else?

Under "Emiley's test" study

### App version

0.1.0(8)

### Device type

Apple

### Operating system version

iOS 18.1.1

pkalita-lbl:
Fixes #213 

The main change here it to use [`toFixed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) on the numeric latitude and longitude values returned by `getCurrentPosition` in order to reduce the number of decimal points.

While I was in the neighborhood, I also added some loading/error state tracking for values (like GPS position) that can be obtained by an async function call. This should help inform the user what's happening if, say, acquiring the GPS position takes more than a few seconds or if it fails entirely.

### DIFF
--- a/src/components/SampleSlotEditModal/SampleSlotEditModal.tsx
+++ b/src/components/SampleSlotEditModal/SampleSlotEditModal.tsx
@@ -318,9 +318,14 @@ const SampleSlotEditModal: React.FC<SampleSlotEditModalProps> = ({
               expand="block"
               onClick={async () => {
                 const position = await Geolocation.getCurrentPosition();
-                handleValueChange(
-                  position.coords.latitude + " " + position.coords.longitude,
-                );
+                // The submission schema allows up to 8 decimal places for lat/lon. The 8th decimal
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
Looks good to me! Thanks for implementing this, and including the explanation in a comment. I'm ready for this to be merged in.

yxu-lanl (approved):
Thanks.

PR #219, opened 2025-01-22
Add additional certificate fingerprint

pkalita-lbl:
The Google Play Console was complaining that it could not verify domain ownership for deep linking. Its suggested fix was to add this additional certificate fingerprint to our `assetlinks.json` file. So that's what I'm doing here 😂


eecavanna (approved):
Thanks for taking care of this.

Issue #187, opened 2024-09-25
Regular expressions are difficult to use as a novice in entering values


sierra-moxon:
### Bug summary

In the env broad scale entry screen, when annotating a sample, the regular expression that I need to follow is very long and hard to examine, understand, and comply with from within the edit screen

### Bug details

Maybe some more granular validation interaction reports would be helpful?

E.g. if I add a term like "ENVO:00000123", I get "Value does not match pattern" and a long paragraph of text about how to select a term from ENVO ("Description") and then "Guidance" is a Pattern that has a lot of special characters that even as a software dev I would paste into an IDE to help me remember. 

Under `Examples,` I see [ENVO:01000033] - does this mean I need brackets around my entry?  Or is ENVO:00000123 too long?  Or is it not a valid term in the ontology?  Or is it an invalid term in a smaller picklist?  Where can I find that picklist in the app?

I tried a few things (as above), but I still got the error so I am not sure how to fill out this field.

### Steps to reproduce

1. Launch app
2. make study
3. make sample
4. fill out env broad scale and try to find a valid value from ENVO. 

### Device type

Apple

### Operating system version

iOS 17.6.1

mslarae13:
Env triad will probably NOT be completed in the field.
So maybe we don't have to worry about it for the app?

pkalita-lbl:
See also: https://github.com/microbiomedata/submission-schema/issues/227

eecavanna:
> the regular expression [...] is very long and hard to examine, understand, and [...]

One way we may be able to mitigate this (brainstorming here) is to apply syntax highlighting to the regular expression. Here's a JavaScript library that can be used to do that: https://slevithan.github.io/regex-colorizer/demo/ (I've never used this library)

eecavanna:
Here's what one of our regexes looks like with the syntax highlighting applied by that library:

![image](https://github.com/user-attachments/assets/9e422ad5-46e8-4a89-b4ec-a57fd2d74435)

I think it took the presentation from "cursed" to "systematically cursed."

pkalita-lbl:
> I think it took the presentation from "cursed" to "systematically cursed."

😂 

Yeah. I was thinking that if you don't already know what a regular expression is it takes from gobbledygook to gobbledygook _in color_.

[...]

PR #224, opened 2025-02-03
Display slot patterns within a dedicated accordion (collapsed by default)

eecavanna:
On this branch, I removed the slot's pattern information (both its `pattern` value and its `structured_pattern.syntax` value) from the general "Guidance" section. Then, I added an accordion—collapsed by default—to the bottom of the slot's help panel, which, when expanded, shows the slot's pattern information (only its `pattern` value, though).

Fixes (mitigates) #187 

Here are some screenshots showing a slot's help panel using the code on this branch:

#### Dark mode (with `ionic:mode=ios`)

image

image

#### Dark mode (with `ionic:mode=md`)

image

image

### Light mode (with `ionic:mode=ios`)

image

image

### Light mode (with `ionic:mode=md`)

image

image


src/components/SchemaSlotHelp/SchemaSlotHelp.tsx
```diff
@@ -9,16 +17,14 @@ const SchemaSlotHelp: React.FC<SchemaSlotHelpProps> = ({
...
   const guidanceParagraphs: string[] = [];
   if (slot.comments && slot.comments.length) {
     guidanceParagraphs.concat(...slot.comments);
   }
-  if (slot.pattern) {
-    guidanceParagraphs.push("Pattern as regular expression: " + slot.pattern);
-  }
-  if (slot.structured_pattern) {
-    guidanceParagraphs.push("Pattern hint: " + slot.structured_pattern.syntax);
-  }
```
pkalita-lbl, line 21: Ah! I didn't even realize that was in there in first place. I still think taking it out completely is the right move until we can be more sure that the `submission-schema` build process is correctly keeping `structured_pattern` and `pattern` in sync.

pkalita-lbl (approved):

PR #226, opened 2025-02-10
Style "Samples" and "Studies" list headers like other section headers

eecavanna:
On this branch, I wrapped the text in the "Samples (N)" header within a `<SectionHeader />` element so that it would look like the section headers higher up on the screen (i.e., be blue and bold). That had some side effects related to margin, which I compensated for with some CSS.

I did the same thing to the "Studies" header on the study list screen. I didn't know whether we would ultimately keep these changes (i.e., the ones on this screen), so I did not consolidate the recurring CSS styles into a single instance yet (e.g., something that could be switched on/off via a prop to the `SectionHeader` component).

Here are before-and-after screenshots showing the changes:

#### Before

- On left: iOS
- On right: Android

image

#### After

- On left: iOS
- On right: Android

image

#### Before

- On left: iOS
- On right: Android

image

#### After

- On left: iOS
- On right: Android

image


src/components/SampleList/SampleList.tsx
```diff
@@ -169,7 +170,11 @@ const SampleList: React.FC<SampleListProps> = ({
   return (
     <>
       <IonListHeader>
-        <IonLabel>Samples {samples && <>({samples.length})</>}</IonLabel>
+        <IonLabel>
+          <SectionHeader className={styles.sectionHeaderWithinListHeader}>
+            <IonLabel>Samples {samples && <>({samples.length})</>}</IonLabel>
```
eecavanna, line 175: Whoops! This `IonLabel`-within-an-`IonLabel`-ness was not intentional. I'll change this.

eecavanna:
If this PR branch gets merged in as is, I'll create a ticket about the "New" button no longer being as distinct (i.e., standing out as much) from the now-blue header text.

eecavanna:
I'm done with this now. Sorry about the thrash (due to (a) mistakes on my part and (b) decisions to refactor based upon how the changes evolved over time).

pkalita-lbl (approved):

PR #228, opened 2025-02-11
Update tour navigation dots and button to use theme's primary color

eecavanna:
On this branch, I made two changes:
1. Update the color of the navigation dots in the tour popup, so they use the app theme's primary color (instead of the NPM package's default color, which was a different shade of blue)
2. Update the "Dismiss" button on the final step of the tour, so it matches the other buttons in the app (specifically, so its background color is the app theme's primary color)

Here are before-and-after screenshots:

### Before

- Left: iOS
- Right: Android

### Light palette

image

> I think I didn't notice the dot's blue shade before because the app theme's primary color was not present in the popup (out of sight, out of mind).

#### Dark palette

image

> I think I didn't notice the dot's blue shade before because the app theme's primary color was not present in the popup (out of sight, out of mind).

### After

- Left: iOS
- Right: Android

#### Light palette

image

#### Dark palette

image

### DIFF
--- a/src/components/AppTourProvider/AppTourProvider.tsx
+++ b/src/components/AppTourProvider/AppTourProvider.tsx
@@ -34,7 +34,11 @@ const AppTourProvider: React.FC<ProviderProps> = ({
         popover: (base) => ({
           ...base,
           borderRadius: 4,
+          // Make the tour text be legible over its background.
           color: "var(--ion-color-primary-contrast)",
+          // Make the navigation dots have the same color as the "Dismiss" button's background.
+          // Reference: https://docs.react.tours/tour/examples#custom-styles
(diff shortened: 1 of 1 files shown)


eecavanna:
I consider the change to the **dot** color to be a "fix". It is an issue that I didn't know existed until I made the button use the app theme's primary color and saw the two shades of blue next to one another.

I consider the change to the **"Dismiss" button** color to be arbitrary (not necessarily a "fix"). In hindsight, I could go either way on it (i.e. keeping it blue—like most, if not all, of the other buttons in the app—or reverting to the light gray background).

pkalita-lbl (approved):
Definitely an improvement!

PR #233, opened 2025-02-18
Let tours be skipped on view enter and started later imperatively

### LINKED ISSUE #231: Tour overlay shows while field selection modal is open
If a user has not yet viewed all app tours and they create a new study, after creating a study they see a tour being presented about the underlying study page while the field visibility modal is open:

Image

pkalita-lbl:
Fixes #231 

The `StudyView` component conditionally launches on modal on data fetch. If it (or any components that it renders) starts a tour on view enter and the modal is opened soon after when the data fetch settles, the result is a tour highlighting an element that is hidden under the modal.

These changes address this by:

* Adding a prop to the `useAppTour` hook that allows a client to turn off initiating the tour on view enter
* Adding a return value from `useAppTour` that includes a function which can imperatively start the tour (or at least attempt to, depending on whether the tour has already been presented)
* Use an extra bit of state in `StudyView` to control whether it is safe to present tours. `StudyView` and its children can observe this state to imperatively start a tour when it's safe to do so.

### DIFF
--- a/src/components/StudyView/StudyView.tsx
+++ b/src/components/StudyView/StudyView.tsx
@@ -47,6 +47,7 @@ const StudyView: React.FC<StudyViewProps> = ({ submissionId }) => {
   const { isOnline } = useNetworkStatus();
   const [modalTemplateVisibleSlots, setModalTemplateVisibleSlots] =
     React.useState<TemplateVisibleSlots | undefined>(undefined);
+  const [toursAllowed, setToursAllowed] = React.useState(false);
 
   const templateVisibleSlots = useMemo(() => {
     const fieldVisibilityInfo: TemplateVisibleSlots[] = [];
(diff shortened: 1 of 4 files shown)


src/components/StudyView/StudyView.tsx
```diff
@@ -268,6 +271,7 @@ const StudyView: React.FC<StudyViewProps> = ({ submissionId }) => {
           )}
 
           <SampleList
+            tourAllowed={toursAllowed}
```
eecavanna, line 274: Not a deal-breaker, from my perspective:

Regarding variable names: when I see the names `tourAllowed`/`toursAllowed`, I think they may contain identifiers of tours that are allowed, as opposed to boolean flags. For that reason, I think I'd prefer prop/variable names that have `is`/`are` in them  (e.g. `isTourAllowed`, `areToursAllowed`).

I don't want to hold up a merge for this. With the TypeScript type information and `React.useState` default value that are present, I (someone reading the code) can resolve the ambiguity.

eecavanna (approved):
LGTM! Thanks for addressing this! I left one comment about boolean prop/variable names. I don't consider it a deal-breaker—just my thoughts as I was reading the code. I'm OK with the branch being merged in as is.

PR #237, opened 2025-02-22
Handle expired refresh tokens

### LINKED ISSUE #118: Handle case where stored refresh token expires
When a user logs in a refresh token is generated for the user and the app stores it in persistent storage. The refresh token used to issue new shorter-lived access tokens. But the refresh token itself will expire after 365 days (by default -- it is configurable per `nmdc-server` environment). When the app detects that the refresh token has expired (as evidenced by  an exchange for a new access token failing), the app needs to revert to a logged-out state and the user will need to log in again.

pkalita-lbl:
Fixes #118 

Apologies in advance for this novel 👇🏻 

### Background

A quick reminder about how authenticated requests to `nmdc-server` API endpoints work and what is currently implemented:

1. When a user logs in they are issued an access token and a refresh token. 
2. The access token is sent with API requests; the refresh token is stored for later. The access token is valid for 24 hours.
3. Once the access token expires, subsequent API requests made with it will result in a `401 Unauthorized` response. 
4. At that point the refresh token needs to be submitted to the `/auth/refresh` endpoint. If the refresh token is valid (it is valid for 365 days), the `/auth/refresh` endpoint will return a new access token. 
5. Then the original API request that generated the `401 Unauthorized` response can be retried with the new access token.

What isn't currently being handled is if the call to `/auth/refresh` in step 4 _also_ returns a `401 Unauthorized` response because the refresh token has expired.

### Summary

The changes here might be considered what our colleague Mark Miller would call a "belt-and-suspenders" approach. Although I'll admit there might be 2 belts and 3 pairs of suspenders here. In other words, there is a lot of redundancy. The reason for the redundancy is that the need for offline support (producing _paused_ API requests) and some oddities about when Android devices resume paused mutations produce a lot of odd edge cases.

At a high level, the approach is two-fold:

* **Preemptively log the user out when their refresh token expires** in order to prevent them from taking any actions that would produce an API request that cannot be fulfilled
* If an API request does trigger a failed token exchange, **catch that specific error and log the user out**

### Details

* A new hook has been added (`src/useLogoutAndRedirect.ts`) to facilitate the log out action. The function provided by this hook differs from the `logout` function provided by the `Store` in that it also does a bit of offline query cleanup (note: just queries! not mutations!), presents a toast message to the user, and returns them to the welcome screen.
* These changes are related to preemptively logging the user out when the refresh token expires:
  * In `src/Store.tsx` we now track when the time that the refresh token will expire. This is done by decoding the JWT refresh token and extracting the `exp` field. 
  * A new component (`RefreshTokenExpirationMonitor`) picks up the refresh token expiration time and, if that time is soon-ish, sets up a timer to perform the logout action.
  * As a convenience to the user (and also as a debugging aid to the developer), the refresh token expiration time is now shown in the User section of the Settings tab (`SettingsUserList`).
* These changes are related to catching and recovering from a case where an API request does somehow go out and fail due to an expired refresh token:
  * In `src/api.ts` we define a new `Error` subclass, `RefreshTokenExchangeError`. This error gets thrown if a call to `/auth/refresh` goes out and fails due to a `401 Unauthorized` response or can't go out due to a missing refresh token.
  * Normally React Query swallows up errors thrown by query and mutation functions so that it can possibly retry them or ultimately invoke the query/mutation `onError` callbacks. These changes to instruct React Query to rethrow the error in the render phase (`throwOnError`) if it is an instance of `RefreshTokenExchangeError`. The error can then be caught by a new error boundary `TokenRefreshErrorBoundary`. Finally when `TokenRefreshErrorBoundary` catches an error it can perform the logout action. 
    > Why all that rigamarole with error boundaries instead of using the query/mutation `onError` callbacks in the first place? First, individual queries don't have an `onError` callback -- only individual mutations do. You can have a _global_ `onError` callback for all queries using `QueryCache` options, but that's outside of the React component tree, so you couldn't utilize `useStore`, `useHistory`, etc from there. Second, even if you could figure out what to do with queries, for mutations you'd still have to remember to put the correct `onError` handler on each individual mutation and that felt, ahem, error prone. For more details see: https://tkdodo.eu/blog/react-query-error-handling
  * The `TokenRefreshErrorBoundary` is wrapped around pages where we make authenticated API requests. Mostly this is handled by adding it to `AuthRoute`. The one exception is the `HomePage`, which does make authenticated API requests if you are logged in, but is also accessible if you are not. So that needed to be wrapped with `TokenRefreshErrorBoundary` in `Router`.
* These changes are about handling mutations that were paused due to being offline. When the app is offline and a mutation is requested, the mutation gets created in a "paused" state and held there for later. There are broadly two cases to consider.
  * If the user makes offline mutations, then closes the app, and reopens the app some time in the future those paused mutations will be rehydrated from the cache. Currently we use the rehydration `onSuccess` callback (`src/QueryClientProvider.tsx`) as an opportunity to resume the paused mutations. However the refresh token might have expired while the app was closed. So these changes add a check to see if the user is logged in before resuming because if they're not logged in those resumed mutation are guaranteed to fail. Since paused mutation might not have been resumed after the cache rehydration, we also want to attempt to resume any paused mutations just after login (`TokenPage.tsx`).
  * The second use case is: the user makes offline mutations, gets logged out due to the refresh token timeout, _keeps the app open_ and then regains network connection. In this case, React Query _really really_ wants to resume those paused mutations immediately after the network reconnect. But if that happens, those mutation requests will go out, fail, and then all that offline work will be lost. The only way I could find to interrupt that process is by making a custom subclass of one of the React Query classes (`NmdcQueryClient`) where I override the `resumePausedMutations` method so that it includes a check for an active refresh token. If it's missing the offline mutations will stay paused until the next login.
* Miscellaneous stuff
  * I initially thought that some features added in a recent version of React Query could help me solve a problem. It turns out that wasn't the case. But I left the upgrade in here because it didn't seem to harm anything.
  * During testing I got sick of having to scroll down to get to the Save button in the `StudyForm` component. So I made it always visible at the bottom of the screen. In addition to solving my own annoyance I think it's a good UX improvement regardless. It did require adding some additional CSS color variables to make it look nice.


eecavanna:
I'll review this on Wednesday.

src/components/SettingsUserList/SettingsUserList.tsx
```diff
@@ -27,6 +27,12 @@ const SettingsUserList: React.FC = () => {
               <p>{loggedInUser.name}</p>
             </IonLabel>
           </IonItem>
+          <IonItem>
+            <IonLabel>
+              <h3>Session expires</h3>
+              <p>{refreshTokenExpiration?.toLocaleString()}</p>
+            </IonLabel>
+          </IonItem>
```
eecavanna, line 35: As I look at this snippet, I am imagining a button here that the user could tap to proactively refresh their session. I don't know yet how that would fit into the whole solution.

src/components/StudyForm/StudyForm.module.css
```diff
@@ -2,3 +2,12 @@
   font-size: var(--nmdc-font-size-xs);
   color: var(--ion-color-danger);
 }
+
+.saveContainer {
```
eecavanna, line 6: Thanks for enhancing the UI in this way!

src/useLogoutAndRedirect.ts
```diff
@@ -0,0 +1,46 @@
...
+import { useQueryClient } from "@tanstack/react-query";
+
+function useLogoutAndRedirect() {
+  const queryClient = useQueryClient();
+  const history = useHistory();
+  const [present] = useIonToast();
+  const { logout, isLoggedIn } = useStore();
+
+  const logoutAndRedirect = useCallback(async () => {
+    // This function may be invoked multiple times by a different components, but the process only
```
eecavanna, line 16: Typo:

```suggestion
    // This function may be invoked multiple times by different components, but the process only
```

eecavanna (approved):
Submitting the review now so that my review comments become public, and then I will re-read the PR description and post a follow-up comment.

eecavanna:
Sorry to say, I think I'm out of time in terms of what I can afford to spend reviewing this PR this week, given unrelated tasks and our plans to submit to the app stores for an initial review by Monday.

I'm OK with this being merged in, given that it does address a previously-unaddressed scenario:
> What isn't currently being handled is if the call to `/auth/refresh` in step 4 also returns a 401 Unauthorized response because the refresh token has expired.

I would like to think more about that scenario, but that isn't something I would be able to do this week. We can revisit this in the future.

src/api.ts
```diff
@@ -253,6 +253,13 @@ export class ApiError extends Error {
   }
 }
 
+export class RefreshTokenExchangeError extends Error {
+  constructor(message: string) {
+    super("Failed to exchange refresh token: " + message);
+    Object.setPrototypeOf(this, RefreshTokenExchangeError.prototype);
```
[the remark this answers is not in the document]
pkalita-lbl: It's a quirk of `Error` subclasses that allows `instanceof` checks to work correctly. See: https://stackoverflow.com/a/76851585.

src/components/SettingsUserList/SettingsUserList.tsx
eecavanna: I added a "Note to self" prefix to this comment. You can resolve this conversation without replying.

src/components/RefreshTokenExpirationMonitor/RefreshTokenExpirationMonitor.tsx
```diff
@@ -0,0 +1,37 @@
...
+    // We don't always set up the timer here because `setTimeout` has a maximum delay of 2^31 - 1
+    // milliseconds, which is about 24.8 days. Many times the expiration will be much further out
+    // than that. We only set up the timer if the expiration is less than 7 days away. This is a
+    // somewhat arbitrary value that is both less than the maximum delay and larger than the amount
+    // of time a user is likely to be continuously using the app.
+    let logoutTimer: NodeJS.Timeout;
+    const msUntilExpiration = refreshTokenExpiration.getTime() - Date.now();
+    if (msUntilExpiration < MAX_DELAY_MS) {
+      logoutTimer = setTimeout(logoutAndRedirect, msUntilExpiration);
+    }
```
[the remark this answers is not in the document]
pkalita-lbl: > But the timeout is still being set for the token expiration time. Did you mean to instead set it to be earlier than that?

No, my rationale is that this is covering the case where the user is actively using the app when the refresh token expires. But in that case they will have an active _access token_, which may be expiring some time between 1 second and 24 hours from now. The user could technically make successful API requests in that period while the access token is active but the refresh token is expired. That gives us a natural buffer period and we don't need to force the log out any amount of time _before_ the refresh token expiration.

> I am wondering whether the countdown pauses while the app is closed

The state of the timer is not persisted when the app closes. The timer just goes away at that point (similar to when a web page is closed).

> when the app is reopened

When the app is reopened, a new timer is setup (if it's within the `MAX_DELAY_MS` threshold) with no connection to the (non-existent) ghosts of timers from sessions past.

src/components/SettingsUserList/SettingsUserList.tsx
pkalita-lbl: > a button here that the user could tap to proactively refresh their session

Getting a new refresh token requires logging out and logging back in. So, no way of doing it automatically per se.

PR #245, opened 2025-03-13
Add contact item to Settings tab

### LINKED ISSUE #242: Add support contact link to About section of Settings tab
We should have a way to contact support in the app itself. I propose that we add a "Contact support" item to the About section of the Settings tab. Tapping on the item should initiate sending an email to `support+fieldnotes@microbiomedata.org` (which gets delivered to the main support inbox).

pkalita-lbl:
Fixes #242 

These changes add a "Contact us" item to the Settings tab. I confirmed that a `mailto:` link does open the default email app (GMail) on an Android simulator. iOS simulators [do not provide an email app](https://discussions.apple.com/thread/3830910?answerId=17959704022&sortBy=rank#17959704022); I trust it will work there as well but we'd need to test on a physical device.

I also changed the existing "Reset tours" item to the default color (it was the primary color). I think the intention was to show that that item was clickable. But we also had other interactive items that were _not_ the primary color. We'll always have a mix of interactive and non-interactive items, but I think I'd rather just be consistent with the colors.

Finally I updated the default value for the support email to be the plussed address we use elsewhere.

### DIFF
--- a/src/components/SettingsAboutList/SettingsAboutList.tsx
+++ b/src/components/SettingsAboutList/SettingsAboutList.tsx
@@ -13,6 +13,17 @@ const SettingsAboutList: React.FC = () => {
           </p>
         </IonLabel>
       </IonItem>
+      <IonItem
+        detail={false}
+        href={`mailto:${config.SUPPORT_EMAIL}`}
+        rel="noopener noreferrer"
+        target="_blank"
+      >
+        <IonLabel>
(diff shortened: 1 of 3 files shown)


eecavanna (approved):
LGTM! Thanks for implementing/testing the email links, and for refining the UI in other ways. I'm OK with this being merged in.

PR #247, opened 2025-03-18
Fix slot selector modal presentation logic

### LINKED ISSUE #246: Template field selector slides up/down repeatedly
### Bug summary

When I create a study having two templates, the first template's field selector works normally, but the second template's field selector goes into an infinite show/hide loop, as shown in the video below.

### Bug details

I'll attach a video as an MP4 instead of a GIF so it does not auto-play. The "flashing" of the video made me dizzy. ⚠ 

https://github.com/user-attachments/assets/ea1c672f-5466-42a3-9f08-613f5809d69e

### Steps to reproduce

Here's how you can reproduce this symptom:

1. Create a study that uses two templates
2. Click "Save" on the field selector for the first template
4. See the symptom on the field selector for the second template

To "escape" from the loop, you can repeatedly tap on the part of the screen where the "Save" button appears (in the upper right). Once you successfully press it, the infinite looping will stop and you will be returned to the "Edit Study" page.

### Anything else?

I have observed this in both the iOS simulator and the web app. In both cases, the backend was the `nmdc-server` in the _development_ environment (not local, and not production).

### App version

1.0.0 (13)

### Device type

Apple (simulated), and Web Browser

### Operating system version

iOS 18.1 (in simulator), and Chrome 134 on macOS

pkalita-lbl:
Fixes #246 

The infinite loop of opening and closing the modal was caused by a bad interaction between the `useEffect` callback that attempts to open the modal when there are templates missing field visibility settings and the `onSuccess` callback called after the submission was updated with new field visibility settings that closes the modal. I believe the order of the `useEffect` call and the `onSuccess` call is not 100% deterministic, which is why there was initially a little confusion about how to reproduce the issue.

The fix here is to remove the `onSuccess` callback entirely. With these changes the `useEffect` callback is responsible for both opening _and_ closing the slot selector modal entirely based on the server data state. There is an added benefit here that the modal will now stay open until each template is processed instead of closing the modal and opening it again for the next template to process. 

Since the modal now stays open, however, it is now necessary to reset the scroll position when the slots it displays change. This is to prevent a confusing interaction where a user scrolls through the list of slots in the modal, taps save (which results in a new set of slots to show), and then they're looking at the middle of the new list.

### DIFF
--- a/src/components/SlotSelectorModal/SlotSelectorModal.tsx
+++ b/src/components/SlotSelectorModal/SlotSelectorModal.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo } from "react";
+import React, { useEffect, useMemo, useRef } from "react";
 import {
   IonButton,
   IonButtons,
@@ -98,6 +98,7 @@ const SlotSelectorModal: React.FC<SlotSelectorModalProps> = ({
   templateName,
 }) => {
   const [selectedSlots, setSelectedSlots] = React.useState<SlotName[]>([]);
+  const contentRef = useRef<HTMLIonContentElement | null>(null);
   const schema = useSubmissionSchema();
 
   const schemaClassName = templateName && TEMPLATES[templateName].schemaClass;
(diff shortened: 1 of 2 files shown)


eecavanna (approved):
LGTM! Thank you for summarizing the problem, solution, and implication about the scroll position.

PR #264, opened 2025-07-11
Use `MutationErrorBoundary` to show error if study creation fails

### LINKED ISSUE #234: App does not provide feedback to user when API request fails
Issue https://github.com/microbiomedata/nmdc-field-notes/issues/230 (now resolved) was about the failure to create a Study. The user would tap the "Save" button at the bottom of the "New Study" screen, but the Study would **not** be created.

In that situation, the app would not show an error message. In other words, the app was not giving feedback to the user about the recently-attempted operation. This may have led to the user wondering whether the app was waiting for them to do something more. That is as opposed to the app displaying a message to the user, saying that an error had occurred.

When we dug into the issue, we found that the app was receiving an HTTP 422 response from the server due to the app not including a required JSON property in its request payload. Since the error was not something that the user could have resolved themselves, the error message could have been [something generic like] "An error occurred."

pkalita-lbl:
Fixes #234 

Does what it says on the tin: use the existing `MutationErrorBoundary` component to show an error message if study creation fails. 

I also did a quick survey of other mutation hooks to see if we were missing any other `MutationErrorBoundary` usage, and I didn't see any. Caveat: some lock/unlock mutations are intentionally not piped to a `MutationErrorBoundary` because if they fail there will a "you don't have the lock" error when the subsequent update mutation goes out and we catch/display _that_ error.

### DIFF
--- a/src/pages/StudyCreatePage/StudyCreatePage.tsx
+++ b/src/pages/StudyCreatePage/StudyCreatePage.tsx
@@ -19,6 +19,7 @@ import { useNetworkStatus } from "../../NetworkStatus";
 import FixedCenteredMessage from "../../components/FixedCenteredMessage/FixedCenteredMessage";
 import useNavigateWithState from "../../useNavigateWithState";
 import { logStudyCreatedEvent } from "../../analytics";
+import MutationErrorBanner from "../../components/MutationErrorBanner/MutationErrorBanner";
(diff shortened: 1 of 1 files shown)


eecavanna (approved):
LGTM! Thanks for implementing this.

PR #268, opened 2025-08-05
Allow creation of test submissions

### LINKED ISSUE #153: Allow user to designate a submission as "test"
This issue is part of the following **meta issue**: https://github.com/microbiomedata/issues/issues/825

In the meta issue, a team member is describing a Submission Portal feature that users can use to indicate that a given submission (of theirs) is not "real"; e.g. they're using it for a workshop, for testing, etc. This issue represents the task of updating the Field Notes app so that users can indicate that same thing via the app.

On a side note, it may be the case that we could omit this functionality from the Field Notes app and, instead, require the user to indicate it downstream, on the Submission Portal. At this point, I don't think this approach has been proposed to the person that created the meta issue.

pkalita-lbl:
Fixes #153 

These changes allow users to indicate via a new checkbox whether the study they are creating is part of a test submission. Once the study is created, the checkbox is still displayed in the study edit form, but it is disabled. In the list of studies displayed on the home screen, a small badge indicates which ones are part of a test submission.

### Details

* Backend support for this feature was already in place. The app just needed to hook into the existing `is_test_submission` field on submission objects. This field was added to the necessary TypeScript interface and data fixtures. Additionally, I cleaned up some leftover interface definitions that are no longer part of submission records (`ContextForm`). 
* The `@ionic/react` dependency is being updated here because the `helperText` prop on the `IonCheckbox` component in version [8.5.0](https://github.com/ionic-team/ionic-framework/releases/tag/v8.5.0).

### Screenshots

| Study Create Form | Study Edit Form | Study List |
| --- | --- | --- |
| image | image | image |


eecavanna (approved):
LGTM!

Side note: I think showing the _full_ help text blurb, even for an existing study, could serve as a reminder for users about why they did/didn't check that box when creating the study (even though they can't change it). This is something we can discuss at a future Field Notes squad meeting or on Slack. I don't want to hold up the merge over it. Thanks for implementing this!

eecavanna:
(OOO, just checking in while pruning iPhone notifications)

PR #269, opened 2025-08-06
Add new setting for keeping the screen on while using the app

### LINKED ISSUE #94: Add setting for keeping screen on while using app
In the GSP town hall, a few participants floated the idea of keeping the screen on and unlocked while using the app so that they don't have to keep unlocking the device in the field.

There is a Capacitor plugin that I believe supports this: https://github.com/capacitor-community/keep-awake. We'd just need to add a switch in the Settings page to enable/disable that.

pkalita-lbl:
Fixes #94 

### Summary

These changes add a new item to the Appearance section of the Settings page for keeping the screen on while the app is open. 

| Setting Off (default) | Setting On |
| --- | --- |
| image | image |

### Details

* The core functionality is provided by a new Capacitor plugin: https://github.com/capacitor-community/keep-awake
* Similar to other settings, the initial state of the setting is pulled from persistent storage. When the setting changes, the new value is used to call the appropriate methods provided by the plugin and then written to persistent state. This is all coordinated in the `<StoreProvider>` component.
* The new `KeepAwakeSwitch` component is then just responsible for checking to see if "keep awake" functionality is supported on the current platform, getting state from the store, and calling the appropriate store function when the user changes the setting.
* While writing tests for the `KeepAwakeSwitch` switch I noticed that mocking wasn't working as described in the Vitest docs. It turns out we were multiple major versions behind in that library. Upgrading to the latest version fixed my mocking issues and didn't seem to cause any problems for existing tests.

### Testing

Aside from the new unit test, I also was able to test this on the Android simulator. You can go into the settings on the simulator and set the screen timeout (Settings -> Display & touch -> Screen timeout). It's off by default, but once I set it to 30 seconds I could confirm the correct behavior of the new setting in the app.

As far as I could tell, the iOS simulator screen never times out and there is no setting to configure it. So I wasn't able to interactively test on iOS. Someone with a physical iOS device might want to do a quick sanity check of this functionality before we do a general release.

### DIFF
--- a/src/components/KeepAwakeSwitch/KeepAwakeSwitch.tsx
+++ b/src/components/KeepAwakeSwitch/KeepAwakeSwitch.tsx
@@ -5,13 +5,13 @@ import { useStore } from "../../Store";
 
 const KeepAwakeSwitch: React.FC = () => {
   const { isKeepAwakeOn, setIsKeepAwakeOn } = useStore();
-  const [keepAwakeSupported, setKeepAwakeSupported] =
+  const [isKeepAwakeSupported, setIsKeepAwakeSupported] =
     React.useState<boolean>(false);
 
   useEffect(() => {
     async function init() {
       const result = await KeepAwake.isSupported();
-      setKeepAwakeSupported(result.isSupported);
+      setIsKeepAwakeSupported(result.isSupported);
     }
     void init();
   }, []);
@@ -19,7 +19,7 @@ const KeepAwakeSwitch: React.FC = () => {
   return (
     <IonToggle
       checked={isKeepAwakeOn}
-      disabled={!keepAwakeSupported}
+      disabled={!isKeepAwakeSupported}
       onIonChange={(e) => {
         setIsKeepAwakeOn(e.detail.checked);
       }}
@@ -27,7 +27,7 @@ const KeepAwakeSwitch: React.FC = () => {
       <IonLabel>
         <h3>Keep Screen On</h3>
         <p>
-          {keepAwakeSupported ? (
+          {isKeepAwakeSupported ? (
             <>May significantly increase battery usage</>
           ) : (
             <>Setting not supported on this device</>


src/components/KeepAwakeSwitch/KeepAwakeSwitch.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { useEffect } from "react";
+import { IonLabel, IonToggle } from "@ionic/react";
+import { KeepAwake } from "@capacitor-community/keep-awake";
+import { useStore } from "../../Store";
+
+const KeepAwakeSwitch: React.FC = () => {
+  const { isKeepAwakeOn, setIsKeepAwakeOn } = useStore();
+  const [keepAwakeSupported, setKeepAwakeSupported] =
```
eecavanna, line 8: I suggest changing `keepAwakeSupported` to `isKeepAwakeSupported` (and similar for the setter function). Not a deal-breaker, in my opinion.

eecavanna (approved):
Thanks for implementing this! (Also, thanks for posting a heads up on Slack a few days ago about this task.)

The PR description and diff look great to me. I left a suggestion about prefixing the name of a Boolean variable with "is" (it's not a deal breaker, from my perspective).

#### iPhone testing

I'm still OOO (final day). If we can get this onto TestFlight in a way that makes it available to me, I can test it on my physical device once I'm back in. I intentionally do not have my work laptop designated as a "trusted device" of my personal iPhone (and vice versa) and so don't have a way to get local builds of the app onto my personal iPhone.

eecavanna (approved):
`is` prefix LGTM — thanks!

eecavanna:
### iPhone testing

I tested this using app version 1.1.0 (15) from TestFlight. 

The feature works on iPhone.

Test sequence:
- Setup: Configure iOS settings such that screen turns off after 30 seconds
- With in-app setting off, set phone on table and watch screen turn off after 30 seconds
- Turn in-app setting on (as shown here)
- With in-app setting on, set phone on table and watch screen remain on after 60 seconds
- Turn in-app setting off, set phone on table and watch screen turn off after 30 seconds
- Teardown: Restore original iOS settings

image

